PR #7, opened 2024-08-30
Implement Chat$stream() using coro

### LINKED ISSUE #5: Add result streaming API
Something like this:

```R
chat <- new_chat()
stream <- chat$chat_stream("Tell me a joke about programming")

# I don't love this style where you simultaneously assign and test:
while(is.null(token <- stream$next_token())) {
  cat(token)
}

# But I don't have an alternative that feels much better
repeat({
  token <- stream$next_token()
  if (is.null(token))
    break

  cat(token)
})
```

jcheng5:
- [x] Sync non-streaming
- [x] Sync streaming
- [x] Async non-streaming
- [x] Async streaming
- [x] Test with tool calling
- [x] Unit tests
- [x] Documentation

Example usage (synchronous):

```r
library(elmer)

chat <- new_chat_openai(quiet = TRUE)


# Sync non-streaming
cat(chat$chat("What are the pros/cons of 1440p vs 4K in a 27\" monitor?"))


# Sync streaming, with coro
coro::loop(for (chunk in chat$stream("What are the pros/cons of 1440p vs 4K in a 27\" monitor?")) {
  cat(chunk)
})
cat("\n")


# Sync streaming, with base R
resp <- chat$stream("What are the pros/cons of 1440p vs 4K in a 27\" monitor?")
repeat {
  chunk <- resp()
  if (coro::is_exhausted(chunk)) {
    cat("\n")
    break
  }
  cat(chunk)
}
```

Example usage (asynchronous):

```r
library(elmer)
library(promises)

chat <- new_chat_async_openai(quiet = TRUE)


# Async non-streaming
chat$chat("What are the pros/cons of 1440p vs 4K in a 27\" monitor?") %...>% cat()


# Async streaming, with coro
coro::async(function() {
  for (chunk in await_each(chat$stream("What are the pros/cons of 1440p vs 4K in a 27\" monitor?"))) {
    cat(chunk)
  }
})()


# Async streaming, with base R 😬
resp <- chat$stream("What are the pros/cons of 1440p vs 4K in a 27\" monitor?")
proceed <- function() {
  resp() |> then(\(chunk) {
    if (coro::is_exhausted(chunk)) {
      cat("\n")
    } else {
      cat(chunk)
      proceed()
    }
  })
}
proceed()
```

Fixes #5


jcheng5:
Points for discussion:

1. Returning coro generators from `Chat$stream()` and coro async generators from `AsyncChat$stream()`
2. The chat() method now yields string content, not a message object. I figured this is what most people would want, and you can peek at `$messages` if you want more detail.
3. The streaming methods yield simple string content. You cannot access the other parts of the delta message. (However, once the streaming is over, you can peek at the message history.) This might be annoying if you want to show progress while tools are being called.
4. In an exchange where the user streams a message, and the chat API invokes tools (possibly multiple rounds of tool invoking), we stream all content without giving the user any indication of message boundaries. This is a little annoying because you end up with the last sentence of one message, butting right up against the first sentence of the next message. For example: "I need to query the database; let me do that now.I've queried the database, here are the results."
5. Pretty much ignoring the existence of choices. If we did want to support them, after each multi-choice response, we'd have to force the caller to choose one before letting them submit another user message.

DESCRIPTION
```diff
@@ -13,6 +13,7 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2
 Imports:
     cli,
+    coro,
```
hadley, line 16: I wonder if we should keep this in suggests? Or do we expect that most folks will be using the coro interface?

R/cat-wordwrap.R
```diff
@@ -0,0 +1,101 @@
...
+cat_word_wrap_impl <- coro::generator(function(con = stdout()) {
+  CAT <- function(...) {
+    cat(..., file = con, sep = "")
+  }
+
+  pos_cursor <- 1
+  buffer <- ""
+
+  while (TRUE) {
+    input <- coro::yield()
```
hadley, line 29: Oh I guess you're using coro here, but presumably this wouldn't be too hard to reimplement as a normal function and not a generator?

R/chat-async.R
```diff
@@ -0,0 +1,271 @@
...
+    #' @description Submit text to the chatbot, and receive a promise that
+    #'   resolves with the response all at once.
+    #' @param text The text to send to the chatbot.
+    #' @returns A promise that resolves to a string (probably Markdown).
+    chat = function(text) {
+      check_string(text)
+
+      # Returns a single message (the final response from the assistant), even if
+      # multiple rounds of back and forth happened.
+      done <- coro::async_collect(private$chat_impl(text, stream = FALSE))
```
hadley, line 106: I guess this is a fundamental part of async, but it kind of sucks that it has to infect the whole stack, meaning that we need two versions of every method and function (sync and async).

R/chat.R
```diff
@@ -25,18 +28,15 @@
 #'   name = "rnorm",
 #'   description = "Drawn numbers from a random normal distribution",
 #'   arguments = list(
-#'     tool_arg(
-#'       "n",
+#'     n = tool_arg(
```
hadley, line 31: This is better!

R/chat-async.R
```diff
@@ -0,0 +1,291 @@
...
+
+    #' @description Submit text to the chatbot, and receive a promise that
+    #'   resolves with the response all at once. If not in quiet mode, the
+    #'   response will be printed to stdout as it is received.
+    #' @param text The text to send to the chatbot.
+    #' @param quiet Whether to emit the response to stdout as it is received. If
+    #'   `NULL`, then the value of `quiet` set when the chat object was created
+    #'   will be used.
+    #' @returns A promise that resolves to a string (probably Markdown).
+    chat = function(text, quiet = NULL) {
```
hadley, line 105: Could this be a `chat_async()` method on the existing class?

Or I wonder if we should more separate what you the methods that you might want to call for interactive experience vs the methods you might call if you're wrapping in another tool? OTOH shiny is like to be the primary use case for async, so we could just design the minimal interface for it?

DESCRIPTION
jcheng5: Let's discuss below

R/cat-wordwrap.R
jcheng5: The moment you use `chat$console()` or `chat$chat(echo=TRUE)` we need coro anyway. And coro has no imports other than rlang. I thought it was worth the dependency.

R/chat-async.R
jcheng5: Yes. This is known as the "what color is your function?" problem. I spent some time considering writing a meta-version of each such method, then being able to "project" sync and async versions of them. It might make sense at some point, but for now it seemed like the bar for contributing to this package would go sky high. (Already higher than I'd like with async, period...)

hadley (approved):

jcheng5:
The R CMD check problems are real but I'll fix them in a separate PR

PR #9, opened 2024-09-09
Changes to quiet mode

jcheng5:
* Change `quiet = TRUE` to `echo = FALSE`
* Default to `echo = FALSE` (i.e. quiet by default)
* `stream`, `chat_async`, and `stream_async` never echo
* Add tests for echoing behavior
* Changed README examples from single code block showing prompt + code, and output, to one code block for code and one code block for output, for easier copy/pasting

### DIFF
--- a/R/chat.R
+++ b/R/chat.R
@@ -14,7 +14,9 @@ NULL
 #'   not supply this directly, but instead set the `OPENAI_API_KEY` environment
 #'   variable.
 #' @param model The model to use for the chat; defaults to GPT-4o mini.
-#' @param quiet If `TRUE` does not print output as its received.
+#' @param echo If `TRUE`, the `chat()` method streams the response to stdout by
+#'   default. (Note that this has no effect on the `stream()`, `chat_async()`,
+#'   and `stream_async()` methods.)
 #' @export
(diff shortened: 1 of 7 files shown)


jcheng5:
Not to be merged until `streaming` branch is merged

R/chat.R
```diff
@@ -14,7 +14,9 @@ NULL
 #'   not supply this directly, but instead set the `OPENAI_API_KEY` environment
 #'   variable.
 #' @param model The model to use for the chat; defaults to GPT-4o mini.
-#' @param quiet If `TRUE` does not print output as its received.
+#' @param echo If `TRUE`, the `chat()` method streams the response to stdout by
```
hadley, line 17: Since it only affects `chat()` it seems weird to make it a class property to me.

hadley (approved):

R/chat.R
jcheng5: I thought so too at first, but for interactive uses (the only place where this will be useful) it seemed annoying to have to remember to do `echo=TRUE` every time?

jcheng5: But I don't feel strongly--happy to take it out if you prefer

PR #11, opened 2024-09-10
Accept messages during construction; add messages accessor

jcheng5:
These changes make it possible to serialize a chat's messages (to disk, database, URL querystring...) and to use those messages to create a new chat.

Since the message history includes the system prompt, `new_chat_openai` will now error if you include both `messages` and `system_prompt` (unless `messages[[1]]$content` and `system_prompt` are the same).


R/chat.R
```diff
@@ -412,22 +448,24 @@ ChatOpenAI <- R6::R6Class("ChatOpenAI",
...
+      system = identity,
+      assistant = cli::col_green
     )
-    cat(cli::rule(message$role, col = color), "\n", sep = "")
-    cat(message$content, "\n", sep = "")
+    cli::cli_rule("{color(message$role)}")
+    # Using cli_text for word wrapping. Passing `"{message$content}"` instead of
+    # `message$content` to avoid evaluation of the (potentially malicious)
+    # content.
+    cli::cli_text("{message$content}")
```
jcheng5, line 461: This change emits the chat history as messages instead of catting to stdout. Is that wrong for a `print` method? I wanted to use `cli_text` for the word wrapping, but can do that via strwrap if we want the history to go to stdout.

R/chat.R
```diff
@@ -49,50 +55,79 @@ NULL
 #'   Briefly explain your work.
 #' ")
 new_chat_openai <- function(system_prompt = NULL,
+                            messages = NULL,
                             base_url = "https://api.openai.com/v1",
                             api_key = openai_key(),
-                            model = "gpt-4o-mini",
+                            model = NA,
```
hadley, line 61: We'd normally make this `NULL` so you can later do `model <- model %||% "gpt-4o-mini"`

R/chat.R
```diff
@@ -9,11 +9,17 @@ NULL
 #' (aka an R function), it also takes care of the tool loop.
 #'
 #' @param system_prompt A system prompt to set the behavior of the assistant.
-#' @param base_url The base URL to the endpoint; the default uses ChatGPT.
+#' @param messages A list of messages to start the chat with (i.e., continuing a
```
hadley, line 12: Might want to briefly the describe its type here?

R/chat.R
```diff
@@ -49,50 +55,79 @@ NULL
...
   check_bool(echo)
 
-  system_prompt <- system_prompt %||%
-    "You are a helpful assistant from New Zealand who is an experienced R programmer"
+  if (is.na(model)) {
+    model <- "gpt-4o-mini"
+  }
+
+  # TODO: Sanity check individual messages?
+  if (!is.null(messages)) {
```
hadley, line 74: A bit prospective maybe, but I'd suggest pulling this out into `check_openai_conversation()`

R/chat.R
```diff
@@ -49,50 +55,79 @@ NULL
...
+        allow_null = TRUE
+      )
+    }
+  }
+
+  if (!is.null(system_prompt)) {
+    if (!is.null(messages) && length(messages) > 0) {
+      if (messages[[1]][["role"]] != "system" ||
+          !identical(messages[[1]][["content"]], system_prompt)) {
+        stop("If both `system_prompt` and `messages` are provided, `messages` must start with the `system_prompt`.")
```
hadley, line 89: I think you could just use `check_exclusive(system_prompt, messages)` and not worry about the more complex check.

R/chat.R
```diff
@@ -49,50 +55,79 @@ NULL
...
+    },
 
-      private$add_message(list(
-        role = "system",
-        content = system_prompt
-      ))
+    #' @description The messages that have been sent and received so far
+    #'   (starting with the system prompt, if any).
+    messages = function() {
+      private$msgs
```
hadley, line 130: Can you not keep this as `private$messges`?

R/chat.R
hadley: It's a litte wrong, but I don't think it's worth trying to do it 100% correctly.

hadley (approved):

R/chat.R
jcheng5: I need an official way to get them out, to serialize existing conversations. And I wanted it publicly read-only for now (although there are scenarios that will demand giving the ability to modify/transform the messages in some way, e.g. RAG, that we can hold off discussing til later)

R/chat.R
jcheng5: Yeah, I was wondering if in this case it was worth preserving the ability to NOT have a `model` argument on the API call. We don't need this today but just in case there is an OpenAI-compatible-ish API in the future that gets popular...

Might be overthinking it...

I guess `NULL` can mean "choose for me" today, and if needed later we could have `NA` mean "explicitly don't include this parameter"? How does that strike you?

tests/testthat/_snaps/chat.md
```diff
@@ -4,6 +4,7 @@
       chat
     Output
       <ChatOpenAI>
```
hadley, line 6: Maybe should add total number of messages here?

R/chat.R
```diff
@@ -412,22 +447,34 @@ ChatOpenAI <- R6::R6Class("ChatOpenAI",
...
+    cli::cli_rule("{color(message$role)}")
+    if (!is.null(message$content)) {
+      # Using cli_text for word wrapping. Passing `"{message$content}"` instead of
+      # `message$content` to avoid evaluation of the (potentially malicious)
+      # content.
+      cli::cli_text("{message$content}")
+    }
+    if (!is.null(message$tool_calls)) {
+      cli::cli_text("Tool calls:")
+      for (tool_call in message$tool_calls) {
```
hadley, line 465: Just convert back to JSON and print it?

jcheng5: Yeah, that was my first instinct...

1. The `tool_calls` can't be printed because it contains random IDs--well, we could, but it makes snapshot testing annoying.
2. We could print each of the tool calls' `function` which would result in `{"name": "rnorm", "arguments": "{\"n\": 5}}` which bothered me because of the noisy double quotes in the key names, but especially, the double-JSON-encoding for `arguments` (because it's actually a string that just happens to be valid JSON).

This way, we get `rnorm({"n": 5})` (which even that, I was tempted to turn into `rnorm(n = 5)`.

tests/testthat/_snaps/chat.md
jcheng5: Like `<ChatOpenAI messages=5>`? Something like that?

R/chat.R
jcheng5: I was just imagining users might want something like:

```r
system_prompt <- "You're a sarcastic idiot"
messages <- maybe_load_from_storage(default = NULL)

chat <- new_chat(system_prompt = system_prompt, messages = messages)
```

Which would work as long as the system prompt hadn't changed.

Your way, the last line would become:

```r
chat <- new_chat(system_prompt = if (is.null(messages)) system_prompt, messages = messages)
```

or maybe the average R programmer would do

```r
if (is.null(messages)) {
  chat <- new_chat(system_prompt = system_prompt)
} else {
  chat <- new_chat(messages = messages)
}
```

Another option would be to say the `system_prompt` is ignored if `length(messages > 0)`?

R/chat.R
jcheng5: I made it `NULL`

R/chat.R
hadley: Sorry, I meant can't the private field name remain as `messages`?

tests/testthat/_snaps/chat.md
hadley: Yeah

R/chat.R
hadley: Hmmmm, good points. We should also contemplate in the context of the claude api where the system prompt is not a message but is a separate parameter to the API?

So for openai, maybe messages should return `private$msgs[1]`, and there's a separate `system_prompt` field?

R/chat.R
hadley: Ah makes sense. For your last point, you could do `call2(funcname, !!!args)` to make a call object which should (I think) print nicely)

R/chat.R
jcheng5: > maybe messages should return `private$msgs[1]`

Do you mean `private$msgs[-1]`?

R/chat.R
jcheng5: No, R6 explicitly checks that names are unique across private, public, and... a third one (protected, maybe?).

R/chat.R
hadley: Oops yes.

R/chat.R
hadley: Oooh, got it, thanks.

tests/testthat/_snaps/chat.md
jcheng5: I added the number of messages _not_ including the system prompt, even though the system prompt is printed right below this line. 🤷‍♂️

PR #16, opened 2024-09-12
Tool helpers

### LINKED ISSUE #15: Semi-automate authoring of register_tool calls
It would be cool to pass a documented package function to a `chat` method, which would use the LLM to turn the function name and helpfile into a correct `$register_tool` call that the user could then copy and paste.

If not a documented function, at least generate as much of the call as possible from the formals, leaving the user to fill in the types and descriptions.

jcheng5:
Closes #15. Currently works with documented package functions, but not with undocumented functions or documented non-package functions.

For undocumented functions, can use the formals to fill out as much as possible and leave the rest (like description) with "TODO", unless the model can infer based on naming or other context clues.

For documented non-package functions... not sure. Use srcrefs to go scrape up the docs?


R/tools.R
```diff
@@ -32,3 +32,95 @@ call_tool <- function(fun, arguments) {
...
+      file_contents <<- c(
+        file_contents,
+        readLines(files, warn = FALSE),
+        "\n"
+      )
+    }
+  }
+
+  op <- options(help_type = "text", pager = fake_pager)
+  on.exit(options(op), add = TRUE)
```
hadley, line 53: There are various slightly more sophisticated techniques we could use here if we wanted to restrict the documentation to just the parameters of interest. But that's something we can easily do later.

R/tools.R
```diff
@@ -32,3 +32,95 @@ call_tool <- function(fun, arguments) {
...
+#'  "gpt-4o", which is highly recommended over "gpt-4o-mini".
+#' @param echo Emit the registration code to the console. Defaults to `TRUE` in
+#'   interactive sessions.
+#'
+#' @return A `register_tool` call that you can copy and paste into your code.
+#'   Returned invisibly if `echo` is `TRUE`.
+#'
+#' @export
+create_tool_metadata <- function(topic, model = "gpt-4o", echo = interactive()) {
+  expr <- rlang::enexpr(topic)
```
hadley, line 95: I think we're importing all of rlang already, or if not, we certainly could.

jcheng5:
`create_tool_metadata` now slurps up roxygen comments (or even non-roxygen comments that immediately precede a function) for functions that don't have a help file.

jcheng5:
I'm thinking of changing `register_tool` to take a single object, so you can have a first-class "function registration object" that you can precompute and, say, attach to a function as an attribute.

From:

```r
chat$register_tool(
  fun = stats::rnorm,
  name = "rnorm",
  description = "Generates random deviates from the normal distribution with specified mean and standard deviation.",
  arguments = list(
    n = tool_arg(
      type = "integer",
      description = "Number of observations. If `length(n) > 1`, the length is taken to be the number required."
    ),
    mean = tool_arg(
      type = "number",
      description = "Mean of the normal distribution. Defaults to 0.",
      required = FALSE
    ),
    sd = tool_arg(
      type = "number",
      description = "Standard deviation of the normal distribution. Defaults to 1.",
      required = FALSE
    )
  )
)
```

to:

```r
chat$register_tool(tool_def(
  fun = stats::rnorm,
  name = "rnorm",
  description = "Generates random deviates from the normal distribution with specified mean and standard deviation.",
  arguments = list(
    n = tool_arg(
      type = "integer",
      description = "Number of observations. If `length(n) > 1`, the length is taken to be the number required."
    ),
    mean = tool_arg(
      type = "number",
      description = "Mean of the normal distribution. Defaults to 0.",
      required = FALSE
    ),
    sd = tool_arg(
      type = "number",
      description = "Standard deviation of the normal distribution. Defaults to 1.",
      required = FALSE
    )
  )
))
```

or actually, I'm not even sure anymore that that's better than a straight port of the OpenAI JSON, given that `create_tool_metadata` is so effective at writing the first version for you:

```r
chat$register_tool(list(
  type = "function",
  function = list(
    name = "get_current_weather",
    description = "Get the current weather in a given location",
    parameters = list(
      type = "object",
      properties = list(
        location = list(
          type = "string",
          description = "The city and state, e.g. San Francisco, CA"
        ),
        unit = list(
          type = "string",
          enum = c("celsius", "fahrenheit")
        )
      ),
      required = c("location")
    )
  )
))
```

jcheng5:
One more idea: defining our own more compact vocabulary.

```r
chat$register_tool(tool_def(
  fun = stats::rnorm,
  name = "rnorm",
  description = "Generates random deviates from the normal distribution with specified mean and standard deviation.",
  arguments = list(
    n = "integer: Number of observations. If `length(n) > 1`, the length is taken to be the number required.",
    mean = "number?: Mean of the normal distribution. Defaults to 0.",
    sd = "number?: Standard deviation of the normal distribution. Defaults to 1."
  )
))
```

This wouldn't work for arguments of type `object` or `array`, I guess we'd need to maintain the more structured way for those cases.

PR #18, opened 2024-09-13
Make it possible to include images in input messages

jcheng5:


R/chat.R
```diff
@@ -535,8 +541,127 @@ print.ChatOpenAI <- function(x, ...) {
...
+#' @export
+#' @examplesIf elmer:::openai_key_exists()
+#' chat <- new_chat_openai(echo = TRUE)
+#' chat$chat(
+#'   list(
+#'     "What do you see in these images?",
+#'     content_image_url("https://www.r-project.org/Rlogo.png"),
+#'     content_image_file(system.file("httr2.png", package = "elmer"))
+#'   )
+#' )
```
hadley, line 553: Do we want to use `...`? e.g.

```suggestion
#' chat$chat(
#'     "What do you see in these images?",
#'     content_image_url("https://www.r-project.org/Rlogo.png"),
#'     content_image_file(system.file("httr2.png", package = "elmer"))
#' )
```

R/chat.R
```diff
@@ -535,8 +541,127 @@ print.ChatOpenAI <- function(x, ...) {
...
+      webp = "image/webp",
+      gif = "image/gif",
+      stop("Unsupported image file extension: ", ext)
+    )
+  }
+
+  if (!is.null(resize)) {
+    # Load the image
+    img <- magick::image_read(path)
+    img <- magick::image_scale(img, "512x512")
```
hadley, line 588: I think that's the correct size for `detail = "low"`, for detail high, it should be 2048x2048

R/chat.R
```diff
@@ -535,8 +541,127 @@ print.ChatOpenAI <- function(x, ...) {
...
+normalize_chat_input <- function(input) {
+  if (is.list(input)) {
+    # If input is already a list, process all elements
+    content <- lapply(input, process_single_input)
+  } else if (is.character(input)) {
+    # If input is a single string, convert it to the required format
+    check_string(input)
+    content <- input
+  } else {
+    stop("Input must be a string or a list")
```
hadley, line 618: You can use `stop_input_type()` here

R/chat.R
```diff
@@ -535,8 +541,127 @@ print.ChatOpenAI <- function(x, ...) {
...
 
 last_message <- function(chat) {
   messages <- chat$messages()
   messages[[length(messages)]]
 }
+
+# Define allowed types - add new types here in the future
+allowed_input_types <- c("text", "image_url")
+
+normalize_chat_input <- function(input) {
```
hadley, line 609: If you want to make this full tidyverse style you'd pass `error_call = caller_env()` along so that error messages can include the user facing function call.

R/chat.R
```diff
@@ -535,8 +541,127 @@ print.ChatOpenAI <- function(x, ...) {
...
+
+    return(item)
+  } else {
+    stop("Each item must be a string or a list")
+  }
+}
+
+format_content <- function(content) {
+  if (is.character(content)) {
+    return(content)
```
hadley, line 659:
```suggestion
   content
```

R/chat.R
```diff
@@ -535,8 +541,127 @@ print.ChatOpenAI <- function(x, ...) {
...
+  } else if (is.list(content)) {
+    return(paste0(lapply(content, function(x) {
+      type <- x[["type"]]
+      value <- x[[type]]
+      paste0("[", type, "]: ", value)
+    }), collapse = "\n"))
+  }
```
hadley, line 666:
```suggestion
    paste0(lapply(content, function(x) {
      type <- x[["type"]]
      value <- x[[type]]
      paste0("[", type, "]: ", value)
    }), collapse = "\n")
  }
```

R/chat.R
jcheng5: 🤦‍♂️  I meant to make it `magick::image_scale(img, resize)`

R/chat.R
jcheng5: That's probably better, yeah. Do I use `rlang::list2(...)`?

R/chat.R
hadley: I think you could still pick for the user.

R/chat.R
jcheng5: Sorry, I tried but could not figure this out. I skimmed https://rlang.r-lib.org/reference/local_error_call.html and then added `error_call = caller_env()` to both `normalize_chat_input` and `process_single_input`, and also added `local_error_call(error_call)` to the top of both function bodies, but after doing that the errors would print with no context at all, instead of the caller as I expected.

R/content.R
```diff
@@ -0,0 +1,161 @@
+#' Encode image content for chat input
```
hadley, line 1: I moved this to a new file, which unfortunately makes my changes harder to see.

R/content.R
```diff
@@ -0,0 +1,161 @@
...
+    content <- paste(input[[1]], collapse = "\n")
+  } else {
+    # Otherwise, process all elements
+    content <- lapply(input, process_single_input, error_call = error_call)
+  }
+
+  list(role = "user", content = content)
+}
+
+process_single_input <- function(item, error_call = caller_env()) {
```
hadley, line 128: Hopefully this gives you a sense of how to use `error_call`.

tests/testthat/_snaps/content.md
```diff
@@ -0,0 +1,58 @@
+# invalid inputs give useful errors
+
+    Code
```
hadley, line 3: We always use snapshot tests for errors since they also include the call, making it easier to skim for issues.

tests/testthat/test-content.R
```diff
@@ -0,0 +1,91 @@
...
+  img_file <- system.file("httr2.png", package = "elmer")
+
+  expect_snapshot(error = TRUE, {
+    content_image_file("DOESNTEXIST")
+    content_image_file(test_path("test-content.R"))
+    content_image_file(img_file, resize = TRUE)
+    content_image_file(img_file, resize = "blah")
+  })
+
+  expect_no_error(content_image_file(img_file))
```
hadley, line 85: You can now use `expect_no_error()`

hadley (approved):

PR #21, opened 2024-09-16
Polish decorator implementation a litte

hadley:
Mostly just so I understand how it works.

### DIFF
--- a/R/coro-utils.R
+++ b/R/coro-utils.R
@@ -10,72 +10,48 @@
 # R6 classes that depend on this will be instantiated at package build time; so
 # the coro generator functions will be burned into the package .Rds file.
 
-# So `R CMD check` doesn't get confused about these variables being used from
-# methods
-utils::globalVariables(c("self", "private", "generator_env", "exits"))
+generators <- new_environment()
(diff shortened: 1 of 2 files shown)


R/coro-utils.R
```diff
@@ -10,72 +10,48 @@
...
 
-  gen <- rlang::inject(
-    function(...) {
-      # Must use elmer::: because the lexical environment of this function is
-      # about to get wrecked by R6
-      elmer:::generators[[!!unique_id]](self, private, ...)
-    }
+  # Must use elmer::: because the lexical environment of this function is
+  # about to get wrecked by R6
+  gen_method <- new_function(args_def,
```
hadley, line 50: This is probably overkill since they're private methods, but this matches the actual argument names rather than relying on `...`.

jcheng5:
Might be tricky merging this with https://github.com/hadley/elmer/pull/17, sorry.

jcheng5:
Is this something you want to come back to?

hadley:
Yeah, I'll re-apply when I next have a minute.

PR #25, opened 2024-09-19
Makes consistent json parsing calls

### LINKED ISSUE #24: Consistent json parsing calls
`jsonlite::fromJSON()` is used in a couple of places instead of `jsonlite::parse_json()`.

The former includes expensive filename detection, which is never needed here.

Will link up a PR for this.

shikokuchuo:
`jsonlite::fromJSON()` -> `jsonlite::parse_json()`

Closes #24.


hadley:
Thanks!

PR #36, opened 2024-09-23
Add model generics

hadley:


R/api.R
```diff
@@ -0,0 +1,142 @@
+chat_request <- new_generic("chat_request", "model")
+
+chat_perform <- function(model,
```
hadley, line 3: I've made this semi-generic but it feels a bit like overkill right now since I don't anticipate claude or gemini needing to override these methods. I might turned them back into regular functions unless you have strong feelings otherwise.

PR #42, opened 2024-09-25
First pass at claude API

hadley:
* [x] Document claude specific params
* [x] Copy OpenAI tests
* [x] Check tool calling support
* [x] Extract text
* [x] Check image upload support — https://github.com/hadley/elmer/issues/54
* [ ] Printing / think about difference in message structure
* [ ] Contemplate [prefilling](https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/prefill-claudes-response)


hadley:
Replaced by #70

PR #95, opened 2024-10-08
Document internals

hadley:
@cpsievert @atheriel I'd love your feedback here. Does this help you understand the key vocab and OO design that powers elmer?


cpsievert:
Yes, this helps, thanks. I had a look at the code yesterday and was a bit confused by the `Turn()` abstraction (also this is my 1st time looking at S7 as well as R for quite some time), but it seems useful now that I more or less understand it as elmer's "standardized" message format. There is also a part of me that would prefer `Message()` over `Turn()`, but that's probably because I'm too aware of Python LLM tooling

atheriel:
Having hung around in the internal for a few days, it does make sense to me, yes. A few observations I have:

- Not all LLMs support images or tool calls. It's unclear how to write a provider that expresses that, other than by throwing a runtime error.
- Not all LLMs providers use API keys. Snowflake and Databricks have OAuth-based credentials that can require refresh flows. I think it's fine to treat these as special cases, though.
- The Snowflake LLM returns some special messages, particularly SQL queries. I used a custom `Content` subclass to express this, but I don't think user turns can contain them and so it's not clear they should be a public API. There's also some weirdness where I'd like to subclass `ContentText` but then override the `@text` property, but that doesn't seem possible with S7's inheritance mechanisms.

hadley:
@cpsievert I did consider Message, but it's a bit confusing, because OpenAI will use multiple messages in the same turn (i.e. you'll have one user message and then potentially multiple tool messages).

@atheriel

* Yeah, I think run-time errors are the best we can do for now. 
* Good point about API keys. Maybe it's better to not pull that up to the super class.
* Your approach in #69 looks good to me. But I think we can do better with the design for accessing text.

PR #98, opened 2024-10-08
Capture token usage in session

hadley:


R/tokens.R
```diff
@@ -0,0 +1,36 @@
+tokens_log <- function(name, tokens) {
+  if (is.null(the$tokens)) {
+    the$tokens <- list()
+  }
+  if (is.null(the$tokens[[name]])) {
+    the$tokens[[name]] <- c(0, 0)
```
jcheng5, line 6: Does this PR assume that prompt and completion (or input and output) are the only token types that exist/will exist? Because OpenAI and Anthropic both have metrics for cached tokens as well.

Not saying we need to account for caching for the tracking to be useful, just that it may make sense to consider the set of tracked metrics to be fairly open-ended.

---

OpenAI:
```
"usage": {
  "prompt_tokens": 2006,
  "completion_tokens": 300,
  "total_tokens": 2306,
  "prompt_tokens_details": {
    "cached_tokens": 1920
  },
  "completion_tokens_details": {
    "reasoning_tokens": 0
  }
}
```

Anthropic:

```
"usage": {"input_tokens":136,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":378}
```

jcheng5 (approved):

R/tokens.R
hadley: At least to begin with, I've leant towards focussing on the token types that everyone has. If you want more specifics for a given back end, you can always do your own iteration over the raw JSON bodies that the assistant turns now possess.

PR #103, opened 2024-10-08
Add some basic prompt helpers

hadley:


R/prompt.R
```diff
@@ -0,0 +1,53 @@
...
+#' topic <- "applies"
+#' prompt(joke)
+prompt <- function(prompt, ..., .envir = parent.frame()) {
+  check_string(prompt)
+
+  if (!dots_named(...)) {
+    cli::cli_abort("All elements of `...` must be named")
+  }
+
+  out <- glue::glue(prompt, ..., .open = "{{", .close = "}}", .envir = .envir)
```
jcheng5, line 31: How about `glue_safe` here? At least by default? Otherwise we're making `.md` a pretty surprising attack vector.

jcheng5:
I can see the usefulness but I'm not convinced this is where we'll ultimately end up.* Therefore I worry that we'll regret giving it such prime real estate (`prompt`). What if you name it not based on the intended use (system prompt) but more literally about what it's doing (template rendering)?

* I've started thinking about where elmer sits relative to the "chain components together" frameworks and am feeling like some of our internals may want to start looking like theirs, and maybe even exposing some extension points, BUT with the crucial difference that we don't lean into the chain-iness as a selling point, especially for people getting started. Like the high level API feels very tailored to the use cases we care about, but you peel back the onion one layer and it's all nicely composable under the hood.

hadley:
How about calling it `interpolate()` for now? It's long and a bit clunky, but still illustrates what it does?

R/prompt.R
hadley: Can you give me a bit more detail on the attack vector that you're thinking of? I was thinking that `prompt` would usually be supplied by the code author, so they're the only ones who get to decide what's evaluated.

jcheng5:
Interpolate sounds great.

PR #113, opened 2024-10-10
Tools schema

### LINKED ISSUE #31: Think about structured data
Either show some examples of using tools + side-effects, or create top-level method to help.

e.g. https://github.com/anthropics/anthropic-cookbook/blob/main/tool_use/extracting_structured_json.ipynb — basic idea  is to define a "print" call and then explicitly request that the model use it. Then you parse the response, either parsing the json or erroring.

hadley:
Todo:

* [x] Think about how to strip off call so openAI can make multiple requests in a row (maybe it should use the native structured data support?). Could add `data_type` to request generic and then each provider can decide what to do with it.
* [x] Force tool usage for claude
* [x] Support optional args for openAI by adding `null` to allowed types
* [x] Add gemini support
* [x] Add some generic test for data extraction
* [x] Add basic spec tests
* [x] More documentation about R <-> json spec
* [x] Move all `as_json_schema` to individual providers.
* [x] Finish off vignette
* [x] Add examples from https://gist.github.com/dannguyen/faaa56cebf30ad51108a9fe4f8db36d8

Fixes #31


jcheng5:
Just getting around to looking at this now. These are just thoughts, nothing that needs addressing before merging this (or ever).

1. Are you going to have `$extract_data_async()`? If not, have you always hated Shiny users or is it a recent development?
2. The vignette is very nice, good examples and clear explanation of the elements of JSON schema.
3. The note about data frames at the end makes me wonder if `type_data_frame` would be useful (along with a helper function for converting array-of-objects to data frame).
4. Is the chat object still usable after performing an `extract_data`? If you're using one of the providers that uses tool calling to implement structured output, do you complete the tool calling loop even though you already got what you wanted?
5. Might be nice to have the equivalent of `create_tool_df` for structured output; you pass in any R object (but probably simple types + named or unnamed list, or data frame), we `str()` or `deparse()` it, and have an LLM do its best at the types.
6. Might be nice to let the user have the option of writing actual JSON schema and importing it, as an alternative to writing `type_xxx()` calls. This would let you use pre-existing JSON schema, JSON schema authoring tools, and JSON schema validators. I've been thinking this for tool calling as well. If the definition is just a string that parses as valid JSON, accept it as-is.

jcheng5 (approved):

hadley:
3. Yeah, not whether that could be fully automated (since an array of objects can always be represented as a data frame) or something you opt-in to.
4. Yes, only claude uses tool calling, and it seems to be ok with never receiving a response.
5. Hmmm, interesting idea. 
6. Oh yeah, that would also be implement.

PR #116, opened 2024-10-11
Fix handling of Workbench-managed Snowflake credentials for Cortex

atheriel:
Ran into this missing `return()` while testing in the Workbench Native App in our dev account. Missed it in #69, oops.


hadley:
Thanks!

PR #119, opened 2024-10-16
Fix usethis function name

hfrick:
The draw of the `use_` prefix is strong but this one is one of the `edit_*()` functions :)


hadley:
Thanks!

PR #125, opened 2024-10-18
Add azure token authentication

### LINKED ISSUE #124: Request to add token-based authorization for Azure
There's another pattern for authenticating to Azure OpenAI that requires a token in addition to the api key. I am (unfortunately) stuck with method. I would love an option to add a bearer token to the Azure API call in addition to the API key. I've implemented this for gptstudio [here](https://github.com/MichelNivard/gptstudio/blob/main/R/service-azure_openai.R).

There's another pattern that uses a Microsoft Entra ID token that Microsoft documents [here](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/managed-identity).

Maybe a `token` arg only for the azure service would do the trick?

JamesHWade:
Adds token-based authentication for azure authentication. Closes #124 

Here's an example of how I get the api calls to work with the changes this PR introduces. 

``` r
library(elmer)
library(AzureAuth)
tenant_id <- Sys.getenv("AZURE_OPENAI_TENANT_ID")
client_id <- Sys.getenv("AZURE_OPENAI_CLIENT_ID")
client_secret <- Sys.getenv("AZURE_OPENAI_CLIENT_SECRET")

retrieve_azure_token <- function() {
  rlang::check_installed("AzureRMR")

  token <- tryCatch(
    {
      AzureRMR::get_azure_login(
        tenant = Sys.getenv("AZURE_OPENAI_TENANT_ID"),
        app = Sys.getenv("AZURE_OPENAI_CLIENT_ID"),
        scopes = ".default"
      )
    },
    error = function(e) NULL
  )

  if (is.null(token)) {
    token <- AzureRMR::create_azure_login(
      tenant = Sys.getenv("AZURE_OPENAI_TENANT_ID"),
      app = Sys.getenv("AZURE_OPENAI_CLIENT_ID"),
      password = Sys.getenv("AZURE_OPENAI_CLIENT_SECRET"),
      host = "https://cognitiveservices.azure.com/",
      scopes = ".default"
    )
  }

  invisible(token$token$credentials$access_token)
}

token <- retrieve_azure_token()

chat <- chat_azure(
  endpoint = Sys.getenv("AZURE_OPENAI_ENDPOINT"), 
  deployment_id = "gpt-4o",
  api_version = Sys.getenv("AZURE_OPENAI_API_VERSION"),
  api_key = Sys.getenv("AZURE_OPENAI_API_KEY"),
  token = token
)

chat$chat("hi")
#> Hello! How can I assist you today?
```

### DIFF
--- a/R/provider-azure.R
+++ b/R/provider-azure.R
@@ -19,7 +19,8 @@ NULL
 #' @param api_key The API key to use for authentication. You generally should
 #'   not supply this directly, but instead set the `AZURE_OPENAI_API_KEY` environment
 #'   variable.
-#' @param token Azure token typically not required for Azure api calls
+#' @param token Azure token for authentication. This is typically not required for
+#' Azure OpenAI API calls, but can be used if your setup requires it.
 #' @inheritParams chat_openai
 #' @inherit chat_openai return
 #' @export
--- a/R/provider-azure.R
+++ b/R/provider-azure.R
@@ -19,6 +19,7 @@ NULL
 #' @param api_key The API key to use for authentication. You generally should
 #'   not supply this directly, but instead set the `AZURE_OPENAI_API_KEY` environment
 #'   variable.
+#' @param token Azure token typically not required for Azure api calls
 #' @inheritParams chat_openai
 #' @inherit chat_openai return
 #' @export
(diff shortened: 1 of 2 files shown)


hadley:
Thanks! Do you want to add the `retrieve_azure_token()` helper too? My goal is to make auth as seamless as possible, but I don't know anything about common Azure patterns so I'd love some help 😄

JamesHWade:
I recommend we put that in a separate PR. It'll take some thinking to make it more general. The way I have it now using a client secret is too specific to my own workflow. Want me to create a separate issue for that so I don't forget to come back to it?

hadley:
Sounds good!

PR #129, opened 2024-10-22
add spec to azure for structured output

JamesHWade:
This copies the approach in provider-openai.R to enable structured output for azure.

Structured outputs have become my most common usage of LLMs lately. Very excited to see these make their way into elmer.


R/provider-azure.R
```diff
@@ -98,11 +98,27 @@ method(chat_request, ProviderAzure) <- function(provider,
 
   # TODO: spec
```
hadley, line 99: Do you mind removing this line too?

hadley:
Thanks!

hadley:
(BTW next on the to do list is some kind of RAG support)

PR #138, opened 2024-10-28
Really take advantage of double-dispatch by using `as_json()` in more places

hadley:
@atheriel do you mind checking that I didn't accidentally break snowflake?


atheriel:
You did not break Snowflake. :+1:

hadley:
@atheriel thanks for checking!

PR #145, opened 2024-11-06
Outsource Snowflake authentication for Cortex to `snowflakeauth`

atheriel:
This is a draft because I'm not actually sure this is a good idea yet and the API of [`snowflakeauth`](https://github.com/atheriel/snowflakeauth) is still WIP.

This commit drops all of the Snowflake-specific authentication logic in favour of using the new `snowflakeauth` package. This has a big advantage in that it works with Snowflake connections and credentials created by the Snowflake Python packages and their CLI.

Calling `chat_cortex()` with no credentials makes our support for these files explicit:

```
Error in `chat_cortex()`:
! The `account` argument is required when ~/.config/snowflake/connections.toml is missing or empty.
ℹ Pass `account` or define a [default] section with an account field in ~/.config/snowflake/connections.toml.
```

However, this is a breaking change in that (1) some environment variables that previously furnished credentials no longer do; and (2) the signature of `chat_cortex()` has changed. The reason to drop the environment variable support is that these specific environment variables are not used by Snowflake's Python tooling -- so I think this is the right choice.

Existing unit tests do cover regressions in the Authorization headers we pass to Snowflake, though I did have to modify them a bit to account for the API changes.


hadley:
Superseded by #183 (I presume)

PR #148, opened 2024-11-07
Implement `chat_vllm()`

### LINKED ISSUE #140: Consider support for vllm-hosted models?
Hi @hadley, thanks for sharing this, really exciting.

Very nice to see support for open models via ollama.  I wonder if you would consider adding support for VLLM-hosted models as well, e.g. see https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#tool-calling-in-the-chat-completion-api

Overall these use the OpenAI API, and indeed we can already do basic chat with these models out of the box by just dropping in a `base_url` to the `openai_chat()`, but it looks like tool use may require a bit more tweaking?  I'm not sure, as it works out-of-the-box with the openai interfaces in langchain, but when I try my tool use examples in elmer i get this error:

```
OpenAIException - Error code: 400 - {'object': 'error', 'message': "[{'type': 'extra_forbidden', 'loc': ('body', 'tools', 0, 'function', 'strict'), 'msg': 'Extra inputs are not permitted', 'input': True}, {'type': 'extra_forbidden', 'loc': ('body', 'tools', 1, 'function', 'strict'), 'msg': 'Extra inputs are not permitted', 'input': True}]", 'type': 'BadRequestError', 'param': None, 'code': 400} 
```

Context: ollama is great for individual use, but expecting every student to have access to a local GPU is often difficult in large classroom settings, and energy-inefficient as well.  A natural solution we've been exploring this semester is to simply host the model on a single node using vllm, which is better at paging many requests than ollama, and serving it to the students as an API, e.g. via LiteLLM, something that we're testing out this semester with the fully open-source function calling model, Gorilla-LLM, developed here at Berkeley.  https://gorilla.cs.berkeley.edu/ .  Would be really keen to do this in R as well.

hadley:
Fixes #140

@cboettig can you please take a look and let me know if this meets your needs?


cboettig:
@hadley thanks much and apologies for the delay.  I can confirm this is working now.  I ran into issues with gorilla, maybe you did too, but was able to test by running other tool-use-enabled models in vllm locally.  

e.g. using `Groq/Llama-3-Groq-8B-Tool-Use`: 

```
#library(elmer)
devtools::load_all()

# vllm is installable as a python module, could have used reticulate
system("vllm serve --dtype=half --enable-auto-tool-choice --tool-call-parser llama3_json  Groq/Llama-3-Groq-8B-Tool-Use")


chat <- chat_vllm(
  base_url = "http://localhost:8000/v1",
  model = "Groq/Llama-3-Groq-8B-Tool-Use",
  api_key = ""
)

tool_rnorm <- tool(
  rnorm,
  "Drawn numbers from a random normal distribution",
  n = type_integer("The number of observations. Must be a positive integer."),
  mean = type_number("The mean value of the distribution."),
  sd = type_number("The standard deviation of the distribution. Must be a non-negative number.")
)
# Then register it
chat$register_tool(tool_rnorm)

# Then ask a question that needs it.
chat$chat("
  Give me five numbers from a random normal distribution of mean 5 and sd 1.
")

```

Groq/Llama-3-Groq-8B-Tool-Use is available of ollama too so should be easy to compare (though sadly no mpb support yet on vllm but fine if you have an nvidia card handy).  

Thanks so much for adding this!  Looking forward to testing out in the classroom!

cboettig:
Maybe this needs to be extended to structured data too?  In calling extract_data() with vllm I'm seeing the errors and it looks like `strict` is still in there, but not sure if it's related.  lemme know if that's worth a separate issue or is an easy extension to this.

hadley:
Oh yeah, I bet I forgot about that.

cboettig:
I think the `no wheels with a matching Python ABI tag` is telling us it can't find any pre-built binary for this hardware (i.e. MBP).  Looks like vllm officially only supports linux (I guess their main target platform is server space)

https://docs.vllm.ai/en/stable/getting_started/installation.html#unsupported-os-build

hadley:
@cboettig in that case, could you run one at some public endpoint I could use to test?

cboettig:
Thanks, tested a bunch of examples.  All the base chat methods are working.  

The open models I tested still seem to be struggling with handling the tool calling and structured data, I'm not entirely sure why.  They don't seem to recognize the API instructions openAI is using, though I think they should?  We can still hack this with vllm and the system prompt, e.g. 

```r
chat$set_system_prompt("The user will supply text input.  From the text, extract the person's name and age and report these using structured JSON")
resp <- chat$chat("
 My name is Susan and I'm 13 years old
")
jsonlite::fromJSON(resp)
```

:joy:


I'm happy with this, seems to do what is needed to support vllm interface

hadley:
@cboettig ok great, I'll add a couple of tests then merge before we release (hopefully in the next couple of days).

PR #160, opened 2024-11-18
Update docs for chat_perplexity

jtrecenti:
Documentation was pointing to Groq. Copied the definition from https://posit-dev.github.io/chatlas/reference/ChatPerplexity.html


hadley:
Oops! Thanks for the fix 😄

PR #165, opened 2024-11-18
Add support for Databricks U2M and M2M OAuth credentials

atheriel:
Use of Databricks personal access tokens are generally discouraged nowadays, the preferred authentication method for Databricks is now OAuth -- either "user to machine", which uses the authorization code flow, or "machine to machine", which is a client credentials flow.

This commit adds support for both. In the process it also adds support for caching and refreshing these credentials correctly across multiple turns.

Unit tests are included for M2M and PAT credentials; U2M credentials are difficult to mock because they require an authorization code flow.

This change also unlocks our ability to use a Databricks service principal to run `elmer`'s own integration tests.

### DIFF
--- a/R/provider-databricks.R
+++ b/R/provider-databricks.R
@@ -9,6 +9,17 @@
 #' outputs. Tool calling support is also very limited at present; too limited
 #' for `elmer`'s tool calling features to work properly at all.
 #'
+#' `chat_databricks()` picks up on ambient Databricks credentials for a subset
+#' of the [Databricks client unified
+#' authentication](https://docs.databricks.com/en/dev-tools/auth/unified-auth.html)
+#' model. Specifically, it supports:
+#'
+#' - Personal access tokens
+#' - Service principals via OAuth (OAuth M2M)
+#' - User account via OAuth (OAuth U2M)
+#' - Authentication via the Databricks CLI
+#' - Posit Workbench-managed credentials
+#'
 #' @family chatbots
 #' @param workspace The URL of a Databricks workspace, e.g.
 #'   `"https://example.cloud.databricks.com"`. Will use the value of the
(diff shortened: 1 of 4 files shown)


R/provider-databricks.R
```diff
@@ -41,7 +53,7 @@ chat_databricks <- function(workspace = databricks_workspace(),
     base_url = workspace,
     model = model,
     extra_args = api_args,
-    api_key = token
+    api_key = token %||% ""
```
hadley, line 56: I find this a bit confusing; I think it makes things a little simpler to reason about if we consistently use `NULL` to mean "use the default".

R/provider-databricks.R
```diff
@@ -147,7 +160,12 @@ databricks_workspace <- function() {
 
 # Try various ways to get Databricks credentials. This implements a subset of
 # the "Databricks client unified authentication" model.
-databricks_token <- function(workspace) {
+databricks_token <- function(workspace = databricks_workspace(), token = "") {
+  # Accept any non-empty user-supplied token.
```
hadley, line 164: I'd suggest a `check_string(token, allow_null = TRUE)` here (along with the appropriate `error_call`).

hadley (approved):
Nice!

R/provider-databricks.R
atheriel: I'd have to change the OpenAI provider for that, it doesn't currently accept a `NULL` `api_key`. Maybe I should define a nullable field for `token` on this provider instead?

R/provider-databricks.R
atheriel: I haven't used argument type checking for internal-only functions so far -- should I start?

R/provider-databricks.R
atheriel: This is what I ended up doing. We no longer equivocate "API key" and "Databricks bearer token". That pattern is used in the Azure provider, too, so things are also more consistent now.

PR #185, opened 2024-11-29
Correcting typo

awynnejo:


hadley:
`README.md` is generated from `README.Rmd` so it needs to be fixed there. It'll be easier for me to do and regenerate the file. Thanks for noticing the problem!

PR #196, opened 2024-12-05
Add support for token object for ProviderAzure. Fixes #195

### LINKED ISSUE #195: Azure chat object fails once token expires
The current method for passing a token to the `ProviderAzure` class passes the token as a string. Once the chat object is created, there is no way to update the token string if the token expires. The consequence is that the entire chat must be re-created, losing all context. 

I am suggesting a change to address this issue by passing a token object from the `AzureAuth` package. The token has built in validate and refresh methods. The chat can now check if the token is valid, and if that fails, attempt to refresh the token.

I wrote up these changes in a pull request. 

I also noticed that the approach to validate the arguments to the `ProviderAzure` class uses the `S7::as_class` function. The issue is that there is not a way to validate an R6 class (what the `AzureAuth` package returns for the token) using the `S7::as_class` function.

I added a workaround in the pull request like this, but I don't love it because it will make the `elmer` package depend on the `AzureAuth` package. I used a blank token to retrieve the class name to ensure that the class "AzureToken" was not hard coded. 

```
prop_azure_token <- function(allow_null = FALSE) {
  force(allow_null)
  ## The call to AzureAuth ensures that the class name of the Azure Token does not change
  class_name = AzureAuth::AzureToken$self$classname
  ## The returned token object is R6 - not currently supported by the S7::as_class()
  ## so I made a new S3 class to use validator
  AzureToken_class <- new_S3_class(class_name)

  new_property(
    class = if (allow_null) NULL | AzureToken_class else AzureToken_class,
    validator = function(value) {
      if (allow_null && is.null(value)) {
        return()
      }

      if (!inherits(value, class_name)) {
        paste0("must be an object of class <", class_name, ">, not ", obj_type_friendly(value), ".")
      }
    }
  )
}
```
I did not add the dependency in the pull request, because maybe there is a better way to do this? I added this code to the pull request, because without it, the change would not work, but any fixes would be very welcome.

SokolovAnatoliy:
This pull request addresses issue #195. The problem solved is that the token passed to the chat object can expire, and then the chat can no longer be continued. 

I propose that the AzureAuth token object be passed instead, which allows for the validate method to determine if the token is expired and the refresh method to refresh the token.

Help is requested on the `prop_azure_token()` function, to eliminate the dependency on the AzureAuth package.

### DIFF
--- a/R/provider-azure.R
+++ b/R/provider-azure.R
@@ -19,8 +19,8 @@ NULL
 #' @param api_key The API key to use for authentication. You generally should
 #'   not supply this directly, but instead set the `AZURE_OPENAI_API_KEY` environment
 #'   variable.
-#' @param token Azure token for authentication. This is typically not required for
-#' Azure OpenAI API calls, but can be used if your setup requires it.
+#' @param token Azure token object of class AzureToken for authentication. This is typically not required for
+#' Azure OpenAI API calls, but can be used if your setup requires it. The token object is retrieved using the AzureAuth package.#' Using the token object ensures a refresh method is available for the token.
 #' @inheritParams chat_openai
 #' @inherit chat_openai return
 #' @export
@@ -58,7 +58,7 @@ ProviderAzure <- new_class(
   parent = ProviderOpenAI,
   properties = list(
     api_key = prop_string(),
-    token = prop_string(allow_null = TRUE),
+    token = prop_azure_token(),
     endpoint = prop_string(),
     api_version = prop_string()
   )
@@ -86,8 +86,18 @@ method(chat_request, ProviderAzure) <- function(provider,
   req <- req_url_query(req, `api-version` = provider@api_version)
   req <- req_headers(req, `api-key` = provider@api_key, .redact = "api-key")
   if (!is.null(provider@token)) {
-    req <- req_auth_bearer_token(req, provider@token)
-  }
+    ## uses the token object method to validate the token (for example if it expired)
+    valid = provider@token$validate()
+    if(!valid ) {
+      # uses the token object method to refresh the token
+      token = provider@token$refresh()
+      access_token = token$credentials$access_token
+    }
+    ## retrieves the actual access token from the object for further use.
+    access_token = provider@token$credentials$access_token
+
+    req <- req_auth_bearer_token(req, access_token)
+    }
   req <- req_retry(req, max_tries = 2)
   req <- req_error(req, body = function(resp) resp_body_json(resp)$message)
 
--- a/R/provider-azure.R
+++ b/R/provider-azure.R
@@ -19,8 +19,8 @@ NULL
 #' @param api_key The API key to use for authentication. You generally should
 #'   not supply this directly, but instead set the `AZURE_OPENAI_API_KEY` environment
 #'   variable.
-#' @param token Azure token object of class AzureToken for authentication. This is typically not required for
-#' Azure OpenAI API calls, but can be used if your setup requires it. The token object is retrieved using the AzureAuth package.#' Using the token object ensures a refresh method is available for the token.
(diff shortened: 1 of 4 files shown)


R/provider-azure.R
```diff
@@ -58,7 +58,7 @@ ProviderAzure <- new_class(
   parent = ProviderOpenAI,
   properties = list(
     api_key = prop_string(),
-    token = prop_string(allow_null = TRUE),
+    token = prop_azure_token(),
```
hadley, line 61: Since we're only using this in a single place, there's no need to pull out into a separate function. And indeed, since we expect that pretty much all `ProviderAzure` objects are created by `chat_azure()`, I'd just do any validation there and not worry about it in the class.

hadley:
But I think we'd prefer not to take a dependency on AzureAuth, so we might need to do this from first principles.

R/provider-azure.R
SokolovAnatoliy: Agreed. That is a good idea. I added a local version of `is_azure_token` from the AzureAuth package to utils, and removed the class validation from utils-S7. Now the validation is done in the `chat_azure` function and is validated as a string in the ProviderAzure class. 

I also renamed the argument token to azure_token, because I thought it was confusing with the term "token" being used for LLM tokens.  Once the access_token is extracted from the object, I called it access_token to make it clear that it was no longer the azure_token object.

atheriel:
@SokolovAnatoliy What form of authentication are you using from `AzureAuth`? Service principals? Managed identities? Azure CLI authentication?

SokolovAnatoliy:
> @SokolovAnatoliy What form of authentication are you using from `AzureAuth`? Service principals? Managed identities? Azure CLI authentication?

@atheriel  - we mainly use Azure Active Directory authentication via authorization_code (thanks to you) and device_code flows. 
These are the first two methods described in the README here:
https://github.com/Azure/AzureAuth

We also use managed identities as well. It depends on the use case.

I am not aware of us using service principals or Azure CLI, but that just may be a limitation of my knowledge based the practices of the users I interact with.

atheriel:
@SokolovAnatoliy If you're using auth code & device code flows with `elmer`, is that for local development/testing? Or have you gotten that wired up in a deployed app of some kind, too?

SokolovAnatoliy:
@atheriel - for `elmer` we are currently using managed identities in Azure, but we do have device_code working in apps, as well as using the auth code flow from the integrations pane in Connect. So I imagine, that both would just work if that was the auth type for LLMs in Azure. 

For us, they all use the exact same structure of creating an R6 object token of the `AzureToken` class. In fact, I convert the token retrieved from Connect Integrations into the R6 `AzureToken`, so that there is no change in experience for the users who utilize such authentications in their work.

atheriel:
@SokolovAnatoliy Thank you, that's very helpful. Great to hear that you're using Connect's OAuth integrations! I'm hoping to get them working with `elmer` automatically soon, too.

JamesHWade:
Any chance improved token handling could make it into main pre-CRAN release? It's exciting to see all the continued progress on the package!

SokolovAnatoliy:
@atheriel - your solution sounds good, and I'd like to check it out, but I just want to comment that my updated code no longer has a dependency on AzureAuth at all.

I updated it to allow `chat_azure` to accept an R6 token object instead of a string as the authentication token. There is a basic check that the object is of class `AzureToken`. That way you to handle the token stuff outside of `chat_azure`. If the token object is no longer valid, the function will attempt to use the token refresh method. In my testing I was able to create a chat with an expired token and confirm that the token refresh method indeed refreshes the token. 

So in the short-medium term, this does not add any dependencies to the ellmer package, but does allow us to use the code with a provided token and keep our chat_azure object from dying every 30 minutes.

atheriel:
See #248.

PR #198, opened 2024-12-05
Adds support for non localhost ollama models. Fixes #197

### LINKED ISSUE #197: Using a hosted ollama model fails due to hard-coded localhost parameters
The current chat_ollama function does not work for ollama models running on a remote machine (non localhost) due to the `base_url` argument not being propagated to the `ollama_models()` and `has_ollama()` functions. 

I propose to add the `base_url` argument to both functions above and pass the `base_url` argument from `chat_ollama()` to both functions.

One other change is to specify just the `base_url` without adding the v (version?) number as the default. The need to have a `/v1` in the `base_url` looks like a requirement for the use of `chat_openai` under the hood? 

My suggestion is that the base_url default is "http://localhost:11434" as suggested by the ollama docs, and then the "v1" is appended only for the `base_url` argument inside the `chat_openai()` function. Perhaps this can somehow be removed or documented, since it is a bit confusing what the `v1` url parameter is doing.

I submitted a pull request to address this issue.

SokolovAnatoliy:
I tried to write up the description in issue #197. In short, this makes it possible to use ollama models hosted on a remote machine.

### DIFF
--- a/R/provider-ollama.R
+++ b/R/provider-ollama.R
@@ -40,7 +40,7 @@ chat_ollama <- function(system_prompt = NULL,
   chat_openai(
     system_prompt = system_prompt,
     turns = turns,
-    base_url = file.path(base_url, "v1"),
+    base_url = file.path(base_url, "v1"), ## the v1 portion of the path is added for openAI compatible API
     api_key = "ollama", # ignored
     model = model,
     seed = seed,
@@ -50,7 +50,8 @@ chat_ollama <- function(system_prompt = NULL,
 }
 
 ollama_models <- function(base_url = "http://localhost:11434") {
-  req <- request(file.path(base_url, "api/tags"))
+  req <- request(base_url)
+  req <- req_url_path(req, "api/tags")
   resp <- req_perform(req)
   json <- resp_body_json(resp)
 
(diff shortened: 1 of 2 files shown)


R/provider-ollama.R
```diff
@@ -40,7 +40,7 @@ chat_ollama <- function(system_prompt = NULL,
   chat_openai(
     system_prompt = system_prompt,
     turns = turns,
-    base_url = base_url,
+    base_url = file.path(base_url, "v1"),
```
hadley, line 43: Would you mind adding a comment here like "for openAI compatible API"?

R/provider-ollama.R
```diff
@@ -49,19 +49,19 @@ chat_ollama <- function(system_prompt = NULL,
   )
 }
 
-ollama_models <- function() {
-  req <- request("http://localhost:11434/api/tags")
+ollama_models <- function(base_url = "http://localhost:11434") {
+  req <- request(file.path(base_url, "api/tags"))
```
hadley, line 53: Would be slightly cleaner to use `req_url_path()` to add `api/tags` to the `base_url`

hadley (commented):
Thanks for working on this!

R/provider-ollama.R
SokolovAnatoliy: Sounds good.

R/provider-ollama.R
SokolovAnatoliy: No problem.

PR #201, opened 2024-12-06
Implement Gemini-specific chunk merging logic

### LINKED ISSUE #199: chat_gemini needs echo = "none" otherwise it doesn't capture full response
I'm trying to capture the response and save it as a variable. Default streaming (text or all) results in only the first word being captured.

```
chat <- chat_gemini(
  model = "gemini-1.5-flash"
)
chat1.gemini1.5flash <- chat$chat(ch1)

chat1.gemini1.5flash
[1] "The"
```

But if I set echo to none then it is successfully captured:

```
chat <- chat_gemini(
  model = "gemini-1.5-flash",
  echo = "none" # echo text or all results in only the beginning of the response being captured
)
chat1.gemini1.5flash <- chat$chat(ch1)

> chat1.gemini1.5flash
[1] "The Moon doesn't have a capital city because it's not a country or a political entity with a government.  It's a natural satellite of Earth.\n"
```

Its only for gemini, no issue capturing the text for openai, claude, or llama

I was asked to add this - Joe says merge_dicts is not working correctly for Gemini, and Google’s implementation is [here](https://github.com/google-gemini/generative-ai-python/blob/b8772ed1424a080911151b354764d76a0e7af2af/google/generativeai/types/generation_types.py#L355-L372)

jcheng5:
Fixes #199

- [ ] Add comments for the merging functions
- [x] `index` is missing from lists, why??
- [x] Implement `merge_safety_ratings`
- [x] Unit tests

Also, something is wrong with tool calling when using this with https://github.com/jcheng5/r-sidebot and you ask it to filter. I think maybe the tool result is being returned with a vestigial text parameter?

```
Warning: Error in req_perform_connection: HTTP 400 Bad Request.
• Unable to submit request because it has an empty text parameter. Add a value to the parameter and try again. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini
```

**Update:** Regarding this HTTP 400 Bad Request, I don't know if it's the right fix but I dealt with it by modifying `method(value_turn, ProviderGemini)` and stripping out parts with nothing but `text=""`. I don't understand why the server was sending me that in the first place.


hadley:
Start at translating go test to R:

```R
r1 <- list(
    candidates = list(
      list(
        index = 2L,
        content = list(role = "model", parts = list(text = "r1 i2")),
        finishReason = 1
      ),
      list(
        index = 0L,
        content = list(role = "model", parts = list(text = "r1 i0")),
        finishReason = 2
      )
    ),
    promptFeedback = list(blockReason = "safety")
  )

  # Create second response
  r2 <- list(
    candidates = list(
      list(
        index = 0L,
        content = list(role = "model", parts = list(text = ";r2 i0")),
        finishReason = 3
      ),
      list(
        index = 1L,
        content = list(role = "model", parts = list(text = ";r2 i1")),
        finishReason = 4
      )
    ),
    promptFeedback = list(blockReason = "other")
  )

  expect_equal(
    merge_gemini_chunks(r1, r2),
    list(
      candidates = list(
        list(
          index = 2L,
          content = list(role = "model", parts = list(text = "r1 i2")),
          finishReason = 1
        ),
        list(
          index = 0L,
          content = list(role = "model", parts = list(text = "r1 i0;r2 i0")),
          finishReason = 3
        )
      ),
      promptFeedback = list(blockReason = "safety")
    )
  )
```

But the level of nesting isn't quite right.

hadley:
* Code should mostly work
* Was tested with sidebot, but running existing tests with `echo = TRUE` should illuminate the problem.
* But gemini chunk size is large, so need to check that's actually
* Then use that to generate some sample inputs to merging function.

PR #206, opened 2024-12-07
copyedit prompt-design vignette: code generation through Token usage

lindbrook:


vignettes/prompt-design.Rmd
```diff
@@ -76,15 +76,15 @@ I'll use `chat_claude()` for this problem because in our experience it does the
...
+When I don't provide a system prompt, I sometimes get answers in different languages or different styles of R code:
 
 ```{r}
 #| label: code-basic
 chat <- chat_claude()
 chat$chat(question)
 ```
 
-I can ensure that I always get R code in a given style by providing a system prompt:
+To ensure that I always get R code in a given style, I provide a system prompt:
```
hadley, line 87: The previous seems a bit clearer to me.

vignettes/prompt-design.Rmd
```diff
@@ -226,7 +226,7 @@ chat <- chat_openai(instruct_json)
 chat$chat(ingredients)
 ```
 
-(I don't know if the colour text, "You're an expert baker who also loves JSON", does anything, but I like to think this helps the LLM get into the right mindset of a very nerdy baker.)
+(I don't know if the text colouring, "You're an expert baker who also loves JSON", does anything, but I like to think this helps the LLM get into the right mindset of a very nerdy baker.)
```
hadley, line 229: I think this changes the meaning, which suggests "colour text" isn't clear. Maybe something like "I don't know if the hint ...` would be better?

vignettes/prompt-design.Rmd
```diff
@@ -251,10 +251,10 @@ chat <- chat_openai(c(instruct_json, instruct_weight))
 chat$chat(ingredients)
 ```
 
-Just providing the examples seems to work remarkably well. But I found it useful to also include description of what the examples are trying to accomplish. I'm not sure if this helps the LLM or not, but it certainly makes it easier for me to understand the organisation of the whole prompt and check that I've covered the key pieces that I'm interested in.
+Just providing the examples seems to work remarkably well. But I found it useful to also include a description of what the examples are trying to accomplish. I'm not sure if this helps the LLM or not, but it certainly makes it easier for me to understand the organisation of the whole prompt and check that I've covered the key pieces I'm interested in.
 
 ```{r}
-#| cacahe: false
+#| cache: false
```
hadley, line 257: Ooops

hadley (commented):
Thanks for your continued help with this. It definitely makes a big difference to the quality of the final product!

vignettes/prompt-design.Rmd
lindbrook: (I don't know if the hint or colour text, "You're an expert baker who also loves JSON", actually does anything, but I like to think it helps the LLM get into the mindset of a very nerdy baker.)

vignettes/prompt-design.Rmd
lindbrook: That reminds me. Up on Line 40 setup chunk, I added an "echo: false" to hide what seemed to me to be a single stray line of code: 

"library(elmer)"

Should that actually be visible (without  "echo: false")? I think I saw the same some other documents/vignettes.

hadley: Yeah, I think it should be visible. In general, we like to make it obvious that we're loading the package at the start of vignettes.

vignettes/prompt-design.Rmd
hadley: Yeah, I like that.

hadley:
Thanks!

PR #222, opened 2024-12-16
Also use GITHUB_PAT env var

### LINKED ISSUE #173: chat_github to retrieve api key using Sys.getenv("GITHUB_API_KEY") instead of using gitcreds function, since gitcreds funtions do not work in Shiny deployment
Currently, chat_github() expects a github credential already setup in the current directory and use gitcreds::gitcreds_get()$password to retrieve the token. I tried to deploy a shiny app using chat_github() with default settings.  It works on my computer and not working after being deployed to shinyapps.io, causing a credential not found error. 

The suggested change is to retrieve github key using Sys.getenv("GITHUB_API_KEY") and have the token placed in the .Renviron file as GITHUB_API_KEY={github_pat_token}. 

There are two benefits making this change:
1. standardize the API management, github API will be placed in the same way as other api keys, e.g., openai api key, inside the .Renviron file.
2. it reduces the complication in shiny deployment, where the deployment environment will likely only have access to .Renviron, not github credentials. 

Thanks

Tony

hadley:
Fixes #173

### DIFF
--- a/R/provider-github.R
+++ b/R/provider-github.R
@@ -13,17 +13,19 @@
 #' @param api_key The API key to use for authentication. You generally should
 #'   not supply this directly, but instead manage your GitHub credentials
 #'   as described in <https://usethis.r-lib.org/articles/git-credentials.html>.
+#'   For headless environments, this will also look in the `GITHUB_PAT`
+#'   env var.
 #' @export
 #' @inheritParams chat_openai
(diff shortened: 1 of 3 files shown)


hadley:
Need to reset `GITHUB_PAT_GITHUB_COM` in test.

PR #232, opened 2024-12-18
Oops, forgot to log body in non-streaming cases

jcheng5:
Repro:

```r
library(elmer)

options(elmer_verbosity = 2)

chat <- chat_claude(model = "claude-3-5-sonnet-latest")
chat$chat("Hello.", echo = FALSE)
chat$chat_async("Hello.")
```

Without this fix, both of these log the requests but not the responses. (Note that in Positron, catting from async handlers doesn't always work, so the `chat_async` verbose output may not appear. Works fine from the terminal and presumably RStudio.)


hadley:
This just confused me!

PR #254, opened 2025-01-15
Pick up on SPARK_CONNECT_USER_AGENT in chat_databricks()

atheriel:
`SPARK_CONNECT_USER_AGENT` is the standard environment variable used to set the user agent in various Databricks and Spark-related projects. This commit adds some detection for when this is set, much like we have in `odbc` and `pysparklyr`.

Unit tests are included.


hadley:
Thanks!

Issue #255, opened 2025-01-15
Add support for Snowflake's "Cortex inference" APIs, backed by Claude


atheriel:
Snowflake just announced that Claude 3.5 Sonnet is now [generally available through Cortex AI](https://www.snowflake.com/en/blog/anthropic-claude-sonnet-cortex-ai/), which is a significant upgrade over any of their existing models. It looks like a lot of their documentation hasn't been updated yet, either, but the starting points would be:

- https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-llm-rest-api#api-reference
- https://docs.snowflake.com/developer-guide/snowflake-rest-api/reference/cortex-inference

We should probably add support for these new "inference" APIs, because I think some folks would be quite excited to route to a high-quality model through Snowflake, much like they're doing today through Azure, Bedrock, or Databricks. These new APIs look to function much more like the other OpenAI-like APIs that `ellmer` supports, too, as opposed to the "Cortex Analyst" APIs that I implemented for `chat_cortex()`.

Maybe we should introduce a new `chat_snowflake()` function for this API and drop the confusing "Cortex" branding of `chat_cortex()` -- or rename it to `chat_cortex_analyst()`, since Snowflake is apparently going to majorly upgrade that API using Claude as well.

hadley:
Renaming proposals sound good to me.

PR #258, opened 2025-01-17
Add a new chat_snowflake() provider

atheriel:
This commit adds `chat_snowflake()` for chatting with models hosted through Snowflake's [Cortex LLM REST API][0]:

    chat <- chat_snowflake()
    chat$chat("Tell me a joke in the form of a SQL query.")

On the backend it looks fairly similar to OpenAI, though it has only the basic textual functionality, and so many advanced ellmer features are not available. I also reused quite a bit of the credential support and utilities from `chat_cortex()`, so this commit also includes some minor refactoring of that provider.

Right now the default model for `chat_snowflake()` is Llama 3.1 70B, but we should change it to Claude 3.5 Sonnet when that gets rolled out more widely; it's only available to customers in the `us-west-1` Snowflake region right now.

Unit tests are included.

Part of #255.

[0]: https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-llm-rest-api


R/provider-cortex.R
```diff
@@ -33,7 +33,8 @@ NULL
 #'   `account`.
 #'
 #' @param account A Snowflake [account identifier](https://docs.snowflake.com/en/user-guide/admin-account-identifier),
-#'   e.g. `"testorg-test_account"`.
+#'   e.g. `"testorg-test_account"`. Defaults to the value of the
+#'   `SNOWFLAKE_ACCOUNT` environment variable.
```
hadley, line 37: Want to cross reference the two snowflake functions?

R/provider-snowflake.R
```diff
@@ -0,0 +1,184 @@
...
+method(chat_request, ProviderSnowflake) <- function(provider,
+                                                    stream = TRUE,
+                                                    turns = list(),
+                                                    tools = list(),
+                                                    type = NULL,
+                                                    extra_args = list()) {
+  if (length(tools) != 0) {
+    cli::cli_abort(
+      "Tool calling is not supported for {.code chat_snowflake()}."
+    )
```
hadley, line 83:
```suggestion
    cli::cli_abort("Tool calling is not supported.", call = quote(chat_snowflake()))
```

Maybe slightly cleaner/simpler?

R/provider-snowflake.R
```diff
@@ -0,0 +1,184 @@
...
+  if (!stream) {
+    cli::cli_abort(
+      "Non-streaming responses are not supported for {.code chat_snowflake()}."
+    )
+  }
+
+  req <- request(provider@base_url)
+  req <- req_url_path_append(req, "/api/v2/cortex/inference:complete")
+  creds <- cortex_credentials(provider@account, provider@credentials)
+  req <- req_headers(req, !!!creds, .redact = "Authorization")
```
hadley, line 99: I wondered about always redacting `Authorization` in httr2. What do you think?

hadley (approved):

R/provider-snowflake.R
atheriel: That kind of makes sense to me, yeah.

R/provider-cortex.R
atheriel: Done.

atheriel:
@hadley I don't have write permissions to merge, if you were waiting for me to.

hadley:
You do now 😄 Feel free to add yourself to the authors too, if you want.

hadley:
You have to accept the invite 😄

Issue #262, opened 2025-01-22
More sophisticated internal APIs for credentials


atheriel:
Right now the complex authentication code for Databricks, Snowflake, and Azure is a bit disorganised and inconsistent (which is my doing, of course). There are also issues with handling Azure and AWS credentials that expire correctly.

Some of this looks like internal mutable state management to me, so I'm thinking of creating an internal `CredentialSource` API using S7 classes and generics that allow us to have sophisticated cloud provider-specific logic around handling headers and expiry. Does that sound like a good idea? I'm also considering whether this could be of use beyond `ellmer`, too, though I'm hesitant to introduce an S7 generic to e.g. `httr2`.

DavZim:
We use Azure OpenAI with access tokens (instead of API keys) and have something along the lines of this when we want to use ellmer.

```R
library(ellmer)

AOAI_access_token <- R6::R6Class(
  "AOAI_access_token",
  public = list(
    access_token = NULL,
    expires_at = NULL,
    initialize = function(token = NULL) {
      if (is.null(token)) {
        self$refresh()
      } else {
        self$access_token <- token$access_token
        self$expires_at <- token$expires_at
      }
    },
    token = function() {
      if (self$expires_at < Sys.time()) self$refresh()
      return(self$access_token)
    },
    refresh = function() {
      response <- httr::POST(
        url = paste0("https://login.microsoftonline.com/",  Sys.getenv("AZURE_TENANT_ID"), "/oauth2/token"),
        body = list(
          "grant_type" = "client_credentials",
          "client_id" = Sys.getenv("AZURE_CLIENT_ID"),
          "client_secret" = Sys.getenv("AZURE_CLIENT_SECRET"),
          "resource" = "https://cognitiveservices.azure.com/"
        )
      )
      
      content <- httr::content(response)
      if (response$status_code != 200) stop(content$error_description)
      
      self$access_token <- content$access_token
      self$expires_at <- Sys.time() + as.numeric(content$expires_in) - 10 # add 10 seconds buffer just to be sure
    },
    print = function() {
      cat("<Azure OpenAI Access Token>:\n")
      cat("Expires at:", format(as.POSIXct(self$expires_at), "%F %T UTC"), "\n")
    }
  )
)

access_token <- AOAI_access_token$new()


chat <- chat_azure(
  endpoint = Sys.getenv("AZURE_OPENAI_ENDPOINT"),
  deployment_id = "gpt-4o",
[...]

PR #263, opened 2025-01-22
Detect credentials from Azure service principals in `chat_azure()`

atheriel:
Inspired by a [recent comment][0], this commit adds support for picking up on Azure service principals, which seem to be a common way to furnish `chat_azure()` with credentials without the need for an API key.

This approach also handles caching and expiry for these tokens, an ongoing source of problems for users.

Unit tests are included. I've also updated the documentation.

[0]: https://github.com/tidyverse/ellmer/issues/262#issuecomment-2606628246

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,9 @@
   API](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-llm-rest-api)
   (#258, @atheriel).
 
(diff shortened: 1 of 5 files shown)


atheriel:
According to https://github.com/tidyverse/ellmer/issues/262#issuecomment-2610803361, this does work, so I think it should be ready for review.

hadley (approved):

PR #266, opened 2025-01-23
Improve support for temporary IAM credentials in chat_bedrock()

### LINKED ISSUE #261: `chat_bedrock()` does not handle AWS credentials that expire
Right now we look up AWS credentials only once, in `chat_bedrock()`: https://github.com/tidyverse/ellmer/blob/bf973c926d2ad8920f837bf8d45dac60b9cf2889/R/provider-bedrock.R#L36

and then use these static credentials to sign requests: https://github.com/tidyverse/ellmer/blob/bf973c926d2ad8920f837bf8d45dac60b9cf2889/R/provider-bedrock.R#L78-L83

This causes problems when using temporary IAM credentials, as is commonly the case when using IAM on Kubernetes or with Workbench-managed credentials (both of which use [`AssumeRoleWithWebIdentity`](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) under the hood).

With the default settings these credentials will expire after an hour, at which point all future requests will fail. You can sidestep this by recreating the chat and passing in the existing Turns, but it seems like we should be handling credential expiry here directly instead.

atheriel:
Many AWS IAM credentials expire, but previously we ignored this by looking up credentials only once, in `chat_bedrock()`. This commit introduces a caching layer that handles expiry and moves credential retrieval closer to request time, instead.

The design is almost identical to httr2's OAuth token caching mechanism, but I had to re-implement various pieces because not all of that API is exported.

(We could probably introduce a `req_aws_credentials()` function to `httr2` itself that would handle this, but that might tie us too closely to the semantics of `paws.common`.)

Unit tests are included.

Closes #261.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,9 @@
   disambiguate it from `chat_snowflake()` (which *also* uses "Cortex") (#275,
   @atheriel).
 
+* `chat_bedrock()` now handles temporary IAM credentials better (#261,
+  @atheriel).
+
 # ellmer 0.1.0
 
 * New `chat_vllm()` to chat with models served by vLLM (#140).
--- a/R/provider-bedrock.R
+++ b/R/provider-bedrock.R
@@ -35,7 +35,8 @@ chat_bedrock <- function(system_prompt = NULL,
                          echo = NULL) {
 
   check_installed("paws.common", "AWS authentication")
-  credentials <- paws_credentials(profile)
+  cache <- aws_creds_cache(profile)
+  credentials <- paws_credentials(profile, cache = cache)
 
   turns <- normalize_turns(turns, system_prompt)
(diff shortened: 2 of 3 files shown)


atheriel:
I haven't been able to a real test of this yet. Should maybe hold off on merging until then.

hadley:
It does feel like maybe httr2 could provide some generic interface for this (i.e. some callback that determines if an error response was due to an out of date token, and another function that updates that token?)

atheriel:
A lot of the pieces are already there with `req_auth_sign()` -- which isn't exported -- but I think much of the value is in implementing caching correctly, which that doesn't help with.

hadley:
I wonder if you can abuse `req_oauth()` to do what you want?

hadley (approved):

PR #269, opened 2025-01-24
Fix the use of `req_error()` for `chat_azure()`

atheriel:
Previously we were trying to access a non-existent field in the response. Now we show both the error code and the message. For example:

    Error in `req_perform_connection()` at elmer/R/httr2.R:36:3:
    ! HTTP 401 Unauthorized.
    • PermissionDenied: Principal does not have access to API/Operation.


hadley:
Came up in #271 too.

PR #273, opened 2025-01-24
Add support for interactive Entra ID authentication to `chat_azure()`

atheriel:
This commit adds support for another major Azure authentication approach: the OAuth authorization code flow, as used by the Azure CLI.

This is a good choice for authentiation during development on desktop, and Microsoft recommends it for Azure OpenAI because it doesn't require storing sensitive long-lived secrets like API keys.

All of this is pretty stock httr2 OAuth stuff, despite the fact that Entra ID has its own... idiosyncrasies. I also went out of the way to add a really specific error message for what I believe to be a common source of problems: misconfiguration of Azure's RBAC. It looks as follows:

    Error in `req_perform_connection()` at elmer/R/httr2.R:36:3:
    ! HTTP 401 Unauthorized.
    • PermissionDenied: Principal does not have access to API/Operation.
    ℹ Your user or service principal likely needs one of the following
      roles: Cognitive Services OpenAI User, Cognitive Services OpenAI
      Contributor, or Cognitive Services Contributor.

I haven't added any unit tests (I don't know how to do so for this kind of interactive OAuth flow), but at least the help documentation has been updated.

### DIFF
--- a/R/provider-azure.R
+++ b/R/provider-azure.R
@@ -18,6 +18,10 @@ NULL
 #' `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, and `AZURE_CLIENT_SECRET` environment
 #' variables are set.
 #'
+#' Finally, in interactive sessions it will also attempt to use Microsoft Entra
+#' ID authentication -- much like the Azure CLI -- if no API key has been
+#' provided.
+#'
 #' @param endpoint Azure OpenAI endpoint url with protocol and hostname, i.e.
 #'  `https://{your-resource-name}.openai.azure.com`. Defaults to using the
 #'   value of the `AZURE_OPENAI_ENDPOINT` envinronment variable.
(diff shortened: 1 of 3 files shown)


R/provider-azure.R
```diff
@@ -123,7 +127,23 @@ method(chat_request, ProviderAzure) <- function(provider,
   req <- req_retry(req, max_tries = 2)
   req <- req_error(req, body = function(resp) {
     error <- resp_body_json(resp)$error
-    paste0(error$code, ": ", error$message)
+    msg <- paste0(error$code, ": ", error$message)
+    # Try to be helpful in the (common) case that the user or service
+    # principal is missing the necessary role.
+    # See: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/role-based-access-control
+    if (error$message == "Principal does not have access to API/Operation.") {
```
hadley, line 134: Maybe use `identical()` here? Otherwise if `message` is every anything other than a string, you'll get a weird error.

R/provider-azure.R
```diff
@@ -123,7 +127,23 @@ method(chat_request, ProviderAzure) <- function(provider,
...
+      msg <- c(
+        "*" = msg,
+        "i" = cli::format_inline(
+          "Your user or service principal likely needs one of the following
+        roles: {.emph Cognitive Services OpenAI User},
+        {.emph Cognitive Services OpenAI Contributor}, or
+        {.emph Cognitive Services Contributor}.",
+          keep_whitespace = FALSE
```
hadley, line 142: Do you need that? I think it will get re-wrapped anyway?

hadley (approved):

R/provider-azure.R
atheriel: In my testing it doesn't get re-wrapped without this :sob:.

atheriel:
Got a real Entra ID RBAC setup working with this today, so I consider it good to go.

hadley:
Feel free to merge once the CI checks pass.

PR #295, opened 2025-02-03
Add api args to Bedrock provider

billsanto:
Added api_args to the Bedrock provider. api_args is a list for parameters such as temperature, top_p, top_k, max_tokens, and stop_sequences. Added validation for each of these parameters. Updated the Roxygen documentation. (Removed the verbose parameter and relevant code/doc comments).


hadley (commented):
Thanks for working on this! A few comments below.

hadley:
Thanks @billsanto, I simplified this a bit more in the light of the fact that we'll invest more in cross-provider parameters in the future.

billsanto:
I see--that is a bit simpler.  Thanks for adding it.

PR #321, opened 2025-02-11
Set connecttimeout with user option (Fixes #318)

CorradoLanera:
The suggestion in https://github.com/tidyverse/ellmer/issues/318#issuecomment-2651431086 by @hadley solves #318, and my reported query (including all the others I have that failed) succeeded.

I've:
- added the corresponding code in the same place with `timeout`
- included the corresponding NEWS item.

PR #322, opened 2025-02-12
Avoid unit tests that snapshot httr2 request objects directly

atheriel:
Instead, test against what usually matters, i.e. the headers, body, or URL.

This should fix test failures in #321.


hadley:
I think these would look better with `str()`

PR #323, opened 2025-02-12
Switch to using `req_dry_run()`

hadley:
Tests won't pass until https://github.com/r-lib/httr2/pull/678 is merged.

@atheriel how does this look to you?


atheriel:
Do you prefer this over #322?

hadley:
@atheriel I mean it prints more attractively 😄 But it's up to you really.

Issue #118, opened 2024-10-12
`create_tool_def` requires OpenAI key


jcheng5:
Either we should document this limitation, or make it possible to use other providers.

jcheng5:
One more option, we could deploy a public Shiny app that lets you paste in a function and you get the ToolDef.

hadley:
Or a plumber API

vorpalvorpal:
Maybe I'm misunderstanding the issue, but I made an identical function that replaced the call to `chat_openai` with `chat_claude`. Couldn't there simply be another parameter that allows the user to select the chat function? If you wanted to get fancy you could have some sort of dictionary that looked up the appropriate chat function based on the model provided. Possibly there could be some environment variable that sets the default chat function and possibly also the default model, and then this and other functions in `elmer` could read that.

PR #325, opened 2025-02-13
Add argument 'chat' to `create_tool_def`

pedrobtz:
- adds argument `chat` to function `create_tool_def` to allow the use of other Chat providers.


hadley:
Fixes #118

R/tools-def-auto.R
```diff
@@ -88,7 +88,9 @@ create_tool_def <- function(topic,
     cli::cli_rule(cli::style_bold("Response"))
   }
 
-  chat <- chat_openai(system_prompt = tool_prompt, model = model, echo = echo)
+  if(!inherits(chat,"Chat"))
+    chat <- chat_openai(system_prompt = get_tool_prompt(), model = model, echo = echo)
```
hadley, line 92: A more common pattern would be

```R
chat <- chat %||% chat_openai(system_prompt = get_tool_prompt(), model = model, echo = echo)
```

I also think it would be easier if we just used the chat as a template, and overrode the existing turns and system prompt, e.g.

```R
chat$clone()
chat$set_system_prompt(get_tool_prompt())
chat$set_turns(list())
```

hadley:
@pedrobtz thanks for getting us started on this. I finished it off by deprecating `model`, unifying the code, and adding a few basic tests.

PR #343, opened 2025-02-27
timestamp when Turns complete

### LINKED ISSUE #337: consider including timestamps in Turns
I'd find it super useful to have timestamps when user and assistant turns completed embedded in Chat/Turn objects. One use case beyond my own that this would make straightforward is calculating tokens/sec. I'd like to use this for an R port of a Python LLM eval framework, [rinspect](https://simonpcouch.github.io/rinspect/index.html).

[Inspect implements this by](https://github.com/UKGovernmentBEIS/inspect_ai/blob/d4458dd9fe0d4d071efdabe4f90cfc1b4818ea76/src/inspect_ai/model/_model.py#L444-L453) recording the time right before a request is made and then another right after it completes, logging the former as the time when the user turn completed and the latter as the time that the assistant turn completed. I've hacked together a [similar implementation](https://github.com/simonpcouch/ellmer/pull/1/files) while testing and it's served my purposes so far, though I haven't verified that timestamps of tool calls are handled reasonably.

I'd rather not implement this in rinspect itself as I'd like for users to be able to supply arbitrary functions wrapping ellmer Chats that define their "solver." With timestamping implemented, rinspect can just derive all of the information it needs from a Chat. Without it, users would have to do their own timestamping inside of those functions. e.g.:

```r
# without built-in timestamping
solver <- function(x, chat) {
  # do some stuff...

  time_start <- Sys.time()
  result <- chat$chat(x)
  time_completed <- Sys.time()
 
  # do some stuff...

  # one timestamp per Turn
  list(result = result, chat = chat, timestamps = list(time_start, time_completed))
}

# with built-in timestamping
solver <- function(x, chat) {
  # do some stuff...

  result <- chat$chat(x)

  # do some stuff...
 
  list(result = result, chat = chat)
}
```

simonpcouch:
Closes #337.

### DIFF
--- a/R/chat.R
+++ b/R/chat.R
@@ -102,7 +102,8 @@ Chat <- R6::R6Class("Chat",
       }
       # Add prompt, if new
       if (is.character(value)) {
-        private$.turns <- c(list(Turn("system", value)), private$.turns)
+        system_turn <- Turn("system", value, completed = NULL)
+        private$.turns <- c(list(system_turn), private$.turns)
       }
       invisible(self)
     },
@@ -399,7 +400,6 @@ Chat <- R6::R6Class("Chat",
         cat_line(format(user_turn), prefix = "> ")
       }
 
-      user_turn@completed <- Sys.time()
       response <- chat_perform(
         provider = private$provider,
         mode = if (stream) "stream" else "value",
@@ -456,7 +456,6 @@ Chat <- R6::R6Class("Chat",
     # If stream = TRUE, yields completion deltas. If stream = FALSE, yields
     # complete assistant turns.
     submit_turns_async = async_generator_method(function(self, private, user_turn, stream, echo, type = NULL) {
-      user_turn@completed <- Sys.time()
       response <- chat_perform(
         provider = private$provider,
         mode = if (stream) "async-stream" else "async-value",
--- a/R/chat.R
+++ b/R/chat.R
@@ -399,6 +399,7 @@ Chat <- R6::R6Class("Chat",
         cat_line(format(user_turn), prefix = "> ")
       }
 
+      user_turn@completed <- Sys.time()
       response <- chat_perform(
         provider = private$provider,
         mode = if (stream) "stream" else "value",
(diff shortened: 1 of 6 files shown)


simonpcouch:
Not able to request review myself, but this should be ready to go @hadley 🏂

tests/testthat/test-chat.R
```diff
@@ -26,7 +26,14 @@ test_that("can retrieve system prompt with last_turn()", {
   expect_equal(chat1$last_turn("system"), NULL)
 
   chat2 <- chat_openai(system_prompt = "You are from New Zealand")
-  expect_equal(chat2$last_turn("system"), Turn("system", "You are from New Zealand"))
+  expect_equal(
+    chat2$last_turn("system"),
+    Turn(
+      "system",
+      "You are from New Zealand",
+      completed = chat2$last_turn("system")@completed
```
hadley, line 34:
```suggestion
      completed = NULL
```

Right?

R/chat.R
```diff
@@ -447,6 +448,7 @@ Chat <- R6::R6Class("Chat",
           cat_line(format(turn), prefix = "< ")
         }
       }
+      turn@completed <- Sys.time()
```
hadley, line 451: Why do you have to update this here instead of relying on the value set by constructor called in `value_turn()` above?

tests/testthat/test-chat.R
simonpcouch: Ah, yup!

hadley:
@simonpcouch I implemented a slightly simpler approach. Can you please take a look and let me know if it makes sense to you?

tests/testthat/test-turns.R
```diff
@@ -1,18 +1,31 @@
+
+
```
simonpcouch, line 2:
```suggestion
```

simonpcouch, line 2:
```suggestion
```

simonpcouch:
Looks great!

PR #345, opened 2025-02-27
Correct function name in tool calling vignette

s-spavound:
In the tool calling vignette there are several references to `create_tool_metadata()` for creating metadata for registering a tool, this appears to be a typo as the function that matches the output in that section is `create_tool_def()`. This PR fixes the vignette to be consistent with the rest of the package.

Please let me know if there is anything else I need to do as I have not contributed to this project before.


hadley:
Thank you!

PR #380, opened 2025-03-24
Add `method(stream_parse, ProviderPerplexity)`

### LINKED ISSUE #379: `method(stream_parse, ellmer::ProviderOpenAI)` causes error in succesful calls to Perplexity API
Using `Chat$chat()` and `Chat$stream()` ends up with a bug when parsing the API response:

``` r
ppx_chat <- ellmer::chat_perplexity(echo = "text", seed = 1)
#> Using model = "llama-3.1-sonar-small-128k-online".

ppx_chat$chat("Respond as brief as possible: Hi")
#> "Hi, how are you?" or "Hi, thanks!" would be brief and professional responses. 
#> If you want to keep it even shorter, "Hi!" is also acceptable in many
#> Error in `method(stream_parse, ellmer::ProviderOpenAI)`:
#> ! Connection closed unexpectedly

ppx_chat$last_turn()@text
#> Error in ppx_chat$last_turn()@text: no applicable method for `@` applied to an object of class "NULL"


ppx_stream <- ellmer::chat_perplexity(echo = "text", seed = 1)
#> Using model = "llama-3.1-sonar-small-128k-online".
stream <- ppx_stream$stream("Respond as brief as possible: Hi")

message_buffer <- ""

coro::loop(for (chunk in stream) {
  message_buffer <- paste0(message_buffer, chunk)
})
#> Error in `method(stream_parse, ellmer::ProviderOpenAI)`:
#> ! Connection closed unexpectedly

ppx_stream$last_turn()@text
#> Error in ppx_stream$last_turn()@text: no applicable method for `@` applied to an object of class "NULL"

message_buffer
#> [1] "\"Hi, how are you?\" or \"Hi, thanks!\" are brief and professional responses. If you want to keep it even shorter, \"Hi!\" is also acceptable in many contexts[1]."
```

<sup>Created on 2025-03-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This is because `stream_parse()` expects a `"[DONE]`" as event data, which is an opinionated implementation of Server-sent Events in OpenAI.

Creating a `method(stream_parse, ProviderPerplexity)` (wich also demands a new `chat_perplexity()`) fixes the issue

<details>

<summary>Implementation</summary>

```r
chat_perplexity <- function(system_prompt = NULL,
                            turns = NULL,
                            base_url = "https://api.perplexity.ai/",
                            api_key = perplexity_key(),
                            model = NULL,
                            seed = NULL,
                            api_args = list(),
                            echo = NULL) {
  
  turns <- ellmer:::normalize_turns(turns, system_prompt)
  model <- ellmer:::set_default(model, "llama-3.1-sonar-small-128k-online")
  echo <- ellmer:::check_echo(echo)
  
  if (ellmer:::is_testing() && is.null(seed)) {
    seed <- seed %||% 1014
  }
  
  provider <- ProviderPerplexity(
    base_url = base_url,
    model = model,
    seed = seed,
    extra_args = api_args,
    api_key = api_key
  )
  
  ellmer:::Chat$new(provider = provider, turns = turns, echo = echo)
}

perplexity_key <- function() {
  ellmer:::key_get("PERPLEXITY_API_KEY")
}

ProviderPerplexity <- S7::new_class(
  "ProviderPerplexity",
  parent = ellmer:::ProviderOpenAI
)

stream_parse <- S7::new_external_generic("ellmer", "stream_parse", "provider")

S7::method(stream_parse, ProviderPerplexity) <- function(provider, event) {
  if (is.null(event) || identical(event$data, "[DONE]")) {
    return(NULL)
  }
  
  jsonlite::parse_json(event$data)
}
```

</details>

``` r
ppx_chat <- chat_perplexity(echo = "text", seed = 1)
#> Using model = "llama-3.1-sonar-small-128k-online".

ppx_chat$chat("Respond as brief as possible: Hi")
#> "Hi, how are you?" or "Hi, thanks for reaching out!" depending on the context 
#> of the message. If you need a more specific response, please provide more 
#> details about the situation.

ppx_chat$last_turn()@text
#> [1] "\"Hi, how are you?\" or \"Hi, thanks for reaching out!\" depending on the context of the message. If you need a more specific response, please provide more details about the situation."


ppx_stream <- chat_perplexity(echo = "text", seed = 1)
#> Using model = "llama-3.1-sonar-small-128k-online".
stream <- ppx_stream$stream("Respond as brief as possible: Hi")

message_buffer <- ""

coro::loop(for (chunk in stream) {
  message_buffer <- paste0(message_buffer, chunk)
})

ppx_stream$last_turn()@text
#> [1] "\"Hi, how are you?\" or \"Hi, thanks!\" depending on the context and relationship with the person. For a more professional response, you could say, \"Hi, I'm doing well, thanks. How about you?\"[1]."

message_buffer
#> [1] "\"Hi, how are you?\" or \"Hi, thanks!\" depending on the context and relationship with the person. For a more professional response, you could say, \"Hi, I'm doing well, thanks. How about you?\"[1].\n"
```

<sup>Created on 2025-03-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I can make a PR for this, since it is already implemented in a branch of my personal fork

calderonsamuel:
# Change description:

- Implemented ProviderPerplexity as child of ProviderOpenAI, with `stream_parse()` method.
- Fixed a bug that generated an error after successful calls to the Perplexity API. Fixes #379 

I noticed that there is no "tests/testthat/test-provider-perplexity.R" file. Would that be a requirement to approve this PR?

### DIFF
--- a/R/provider-perplexity.R
+++ b/R/provider-perplexity.R
@@ -32,20 +32,38 @@ chat_perplexity <- function(system_prompt = NULL,
                             api_args = list(),
                             echo = NULL) {
 
+  turns <- normalize_turns(turns, system_prompt)
   model <- set_default(model, "llama-3.1-sonar-small-128k-online")
+  echo <- check_echo(echo)
 
-  chat_openai(
-    system_prompt = system_prompt,
-    turns = turns,
+  if (is_testing() && is.null(seed)) {
+    seed <- seed %||% 1014
+  }
+
+  provider <- ProviderPerplexity(
     base_url = base_url,
-    api_key = api_key,
     model = model,
     seed = seed,
-    api_args = api_args,
-    echo = echo
+    extra_args = api_args,
+    api_key = api_key
   )
+
+  Chat$new(provider = provider, turns = turns, echo = echo)
 }
 
 perplexity_key <- function() {
   key_get("PERPLEXITY_API_KEY")
+}
+
+ProviderPerplexity <- new_class(
+  "ProviderPerplexity",
+  parent = ProviderOpenAI
+)
+
+method(stream_parse, ProviderPerplexity) <- function(provider, event) {
+  if (is.null(event) || identical(event$data, "[DONE]")) {
+    return(NULL)
+  }
+
+  jsonlite::parse_json(event$data)
 }


R/provider-perplexity.R
```diff
@@ -32,20 +32,38 @@ chat_perplexity <- function(system_prompt = NULL,
...
   key_get("PERPLEXITY_API_KEY")
 }
+
+ProviderPerplexity <- new_class(
+  "ProviderPerplexity",
+  parent = ProviderOpenAI
+)
+
+method(stream_parse, ProviderPerplexity) <- function(provider, event) {
+  if (is.null(event) || identical(event$data, "[DONE]")) {
```
hadley, line 64: I do wonder if we should just make this change in the `ProviderOpenAI` method, since it seems to affect a lot of backends.

calderonsamuel: Tough call! I guess the `data: [DONE]` event makes sense for servers that wouldn't stop listening unless they get that specific event. The OpenAI "/chat/completions" endpoint provides the [`stop`](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stop) parameter that might have use for it

hadley: I assume the python openAI SDK must look for either `"[DONE]"` or (the python equivalent of) `NULL` since otherwise these providers wouldn't claim to be openAI compatible.

calderonsamuel: That seems to be accurate: https://github.com/openai/openai-python/blob/f66d2e6fdc51c4528c99bb25a8fbca6f9b9b872d/src/openai/_streaming.py#L52-L101

I wasn't aware of how much error handling can be added to the process. 

Does this mean that it makes more sense to make the change in `ProviderOpenAI`?

hadley: Yeah, I think so. And then remove any other custom methods where we've done this before. It would also be useful to add that link as a comment. (Thanks for finding it!)

calderonsamuel: All right, I can do that. However, I don't think I'll be able to until the weekend. May I suggest to merge this PR (since it fixes the issue in-scope) and create a new issue for "Stop generating error when event data is NULL in `method(stream_parse, ProviderOpenAI)`"?

hadley: I'd rather not add code that we then have to remove soon after.  But I'm happy to do the fixes and should (🤞) be able to get to it on Thursday or Friday.

PR #388, opened 2025-03-27
Update URLs to OpenAI docs in content-image.R

michaelgrund:
Some of the URLs pointing to the OpenAI docs are not valid anymore. This PR updates the corresponding URLs.


hadley:
Thank you!

PR #391, opened 2025-03-27
Correctly implement citation merging for gemini

### LINKED ISSUE #358: Stream merging fails with gemini
I encountered an error when attempting to create a tool definition for the `get_pums` function from the `tidycensus` package using the `create_tool_def` function from `ellmer`.

**Reproduction steps:**

1.  Load the `tidycensus` and `ellmer` libraries:

    ```r
    library(tidycensus)
    library(ellmer)
    ```
2.  Create a `chat_gemini` object:

    ```r
    chat <- chat_gemini("gemini-2.0-flash")
    ```
3.  Attempt to create a tool definition for `get_pums`:

    ```r
    create_tool_def(get_pums, chat)
    ```

**Expected behavior:**

I expected `create_tool_def` to successfully generate a tool definition for the `get_pums` function, allowing it to be used within the `ellmer` framework.

**Actual behavior:**

Instead of creating the tool definition, the code produced the following response/error:

```r
tool(
  tidycensus::get_pums,
  "Load data from the American Community Survey Public Use Microdata Series API",
  variables = type_array(
    "A vector of variables from the PUMS API. Use `View(pums_variables)` to browse variable options.",
    items = type_string()
  ),
  state = type_string(
    "A state, or vector of states, for which you would like to request data. The entire US can be requested with `state = \"all\"` - though be 
patient with the data download!",
    required = FALSE # TODO: could also be a vector, unclear how to Error in merge_func(left, right, path) : is.list(left) is not TRUE
Error during wrapup: 'S4SXP': should not happen - please report
Error: no more error handlers available (recursive errors?); invoking 'abort' restart
```

**Additional Information:**

- I've tried create_tool_def with other functions (e.g. sum) and it seems to work, so it seems to be related to something specific about the tidycensus::get_pums function.

hadley:
Fixes #358

@jcheng5 thought you might want to review since you wrote this code originally.


jcheng5:
I just realized I have this uncommitted change sitting in my local ellmer repo:

```r
diff --git a/R/provider-gemini.R b/R/provider-gemini.R
index f90507b..be1de8b 100644
--- a/R/provider-gemini.R
+++ b/R/provider-gemini.R
@@ -370,7 +370,15 @@ merge_optional <- function(merge_func) {
 merge_objects <- function(...) {
   spec <- list(...)
   function(left, right, path = NULL) {
+    # This can happen with the finishReason is something other than STOP
+    if (is.null(right)) {
+      return(left)
+    }
+
     # cat(paste(collapse = "", path), "\n")
+    if (!is.list(right)) {
+      message("right is not a list in ", path, "!")
+    }
     stopifnot(is.list(left), is.list(right), all(nzchar(names(spec))))
     mapply(
       names(spec),
```

Note that it is in `merge_objects`, not `merge_optional`, and in the case of citations `merge_optional` passes through to `merge_objects`.

1. Should we add `if (is.null(right)) { return(left} }` to `merge_objects`? I think definitely yes, since there's a comment there explaining that it can happen.
2. Should we also add `if (is.null(left)) { return(right) }` to `merge_objects`? Seems like yes? In theory if we don't believe this could happen we should error and check our assumptions when it does happen, but in practice, I feel like the fix will always turn out to be `if (is.null(left)) { return(right) }` anyway.
3. If we have 1 and 2, do we still need/want `merge_optional` to do the early returns it's doing in this PR? I could go either way. In defense of not having the early returns, `merge_optional` is designed to only protect against the case where the element NEVER appears--it has no opinion on how you merge a null and a non-null value. But again, in practice, are we ever going to not just return the non-null value?

hadley:
Since you have a better sense of how this should flow, I think we should do 1 and 2 and revert the changes to `merge_optional()`.

Issue #340, opened 2025-02-25
Add content types for reasoning


hadley:
At least for claude and openai

hadley:
@simonpcouch there's sufficient variation here that I don't think we should put in `params()`.

wklimowicz:
I think it's the same issue with `chat_databricks` too, for the `databricks-gpt-oss-120b` model:

```r
library(ellmer)

chat <- chat_databricks(model = "databricks-gpt-oss-120b")

chat$chat("test")
```

<details>
<summary>Output</summary>

```
list(type = "reasoning", summary = list(list(type = 
"summary_text", text = "The")))list(type = "reasoning", 
summary = list(list(type = "summary_text", text = " 
user")))list(type = "reasoning", summary = list(list(type =
"summary_text", text = " just")))list(type = "reasoning", 
summary = list(list(type = "summary_text", text = " 
wrote")))list(type = "reasoning", summary = list(list(type 
= "summary_text", text = " \"")))list(type = "reasoning", 
summary = list(list(type = "summary_text", text = 
"test")))list(type = "reasoning", summary = list(list(type 
= "summary_text", text = "\".")))list(type = "reasoning", 
summary = list(list(type = "summary_text", text = " 
Lik")))list(type = "reasoning", summary = list(list(type = 
"summary_text", text = "ely")))list(type = "reasoning", 
summary = list(list(type = "summary_text", text = " 
wants")))list(type = "reasoning", summary = list(list(type 
= "summary_text", text = " a")))list(type = "reasoning", 
summary = list(list(type = "summary_text", text = " 
response")))list(type = "reasoning", summary = 
list(list(type = "summary_text", text = ".")))list(type = 
"reasoning", summary = list(list(type = "summary_text", 
text = " Could")))list(type = "reasoning", summary = 
list(list(type = "summary_text", text = " be")))list(type =
"reasoning", summary = list(list(type = "summary_text", 
text = " just")))list(type = "reasoning", summary = 
list(list(type = "summary_text", text = " 
testing")))list(type = "reasoning", summary = 
list(list(type = "summary_text", text = ".")))list(type = 
"reasoning", summary = list(list(type = "summary_text", 
text = " Respond")))list(type = "reasoning", summary = 
list(list(type = "summary_text", text = " 
politely")))list(type = "reasoning", summary = 
list(list(type = "summary_text", text = ".")))Hello! How 
can I assist you today?

```

</details>

[...]

PR #394, opened 2025-03-31
Implement `params()` wrapper

### LINKED ISSUE #280: Provide common interface to common parameters like `temperature`
Probably using an options object.

And then experiment with setting `temperature` to 0 to make tests more reliable.

hadley:
Goal is to provide common interface for common cross-model parameters.

Fixes #280

@simonpcouch would you mind taking a look at the proposed API? In particular, do you think warning on unsupported parameters is the right choice? I was thinking of evaluation here, since I thought you might want to use the same set of parameters across multiple providers.


R/params.R
```diff
@@ -0,0 +1,74 @@
...
+#' across multiple providers.
+#'
+#' @param temperature Temperature of the sampling distribution.
+#' @param top_p The cumulative probability for token selection.
+#' @param top_k The number of highest probability vocabulary tokens to keep.
+#' @param frequency_penalty Frequency penalty for generated tokens.
+#' @param presence_penalty Presence penalty for generated tokens.
+#' @param seed Seed for random number generator.
+#' @param max_tokens Maximum number of tokens to generate.
+#' @param logprobs Include the log probabilities in the output?
```
simonpcouch, line 19:
```suggestion
#' @param log_probs Include the log probabilities in the output?
```

?

tests/testthat/test-params.R
```diff
@@ -0,0 +1,61 @@
+# params -------------------------------------------------------------------
+
+test_that("NULL values are stripped", {
+  expect_equal(params(), set_names(list()))
+})
+
+test_that("checks its inputs", {
+  expect_snapshot(error = TRUE, {
```
simonpcouch, line 8: Have never thought to intentionally trigger many errors inside of one `expect_snapshot()` call😮

hadley:
I think reasoning is out of scope for this PR, because I think it will require some changes to support thinking tokens first, i.e. #340.

R/params.R
hadley: Yeah, good idea.

PR #405, opened 2025-04-01
Provider a default role for OpenAI messages

hadley:
Part of #161

@atheriel this brings ellmer closer in line with the ChatGPT sdk, but can you please double check it still works for snowflake?


atheriel:
Confirmed that `chat_snowflake()` still works with this change. :white_check_mark:

PR #414, opened 2025-04-02
copyedit vignette:Tool/function calling

lindbrook:


lindbrook:
caller v. user. In the text and diagrams, the use of "user" reads like "you" have to do things manually (e.g., respond to the agent). So maybe just always using "caller" might be better - it's your computer.

This is reinforced with the stick figure user in the diagrams. Maybe a laptop/computer icon?

lindbrook:
A time line arrow from top to bottom might be helpful in the diagrams. It shows that the diagram describes a flow of events.

Labeling the time line (e.g., t1, t2, etc) for each event might also be useful. The uneven vertical spacing (due to the text) makes it seem like there's a difference in execution time when you really are just showing the order of events.

lindbrook:
Consider presenting the two graphs (the "wrong" but intuitive graph and the "right" but unintuitive one) right after the first sentence/paragraph. Then go into what is now the second paragraph - the detailed, illuminating explanation of how things actually work.

FWIW, kind of begs the question of why, other than relieving "them" of the computational burden, it work this way.

PR #422, opened 2025-04-03
Add `echo` as an option (untested)

dylanpieper:
Hi Hadley,

This PR adds `ellmer_echo` as an option. My motivation is to improve batching in `hellmer` by setting echo to `none`. This method allows users to safely pass a chat provider object to the batch processor rather than a function where the processor sets `echo` behind the scenes. I don't like the function method from a UI perspective. It's not very R-like. Right now, if a user supplies a chat provider object, they would have to set echo to `none` themselves.

This is a small change. I didn't write any tests, but here is a reprex demonstrating the behavior:

``` r
library(ellmer)

prompt <- "Explain y = mx + b in one sentence"

options(ellmer_echo = "all") 
cht1 <- chat_openai() 
#> Using model = "gpt-4o".
resp1 <- cht1$chat(prompt)
#> > Explain y = mx + b in one sentence
#> < The equation \( y = mx + b \) represents the equation of a straight line in 
#> < slope-intercept form, where \( m \) is the slope (indicating the steepness 
#> < and direction of the line) and \( b \) is the y-intercept (the point where 
#> < the line crosses the y-axis).
#> <

options(ellmer_echo = "text") 
cht2 <- chat_openai() 
#> Using model = "gpt-4o".
resp2 <- cht2$chat(prompt)
#> The equation \( y = mx + b \) represents the equation of a straight line in a 
#> two-dimensional Cartesian coordinate system, where \( y \) is the dependent 
#> variable, \( m \) is the slope of the line, \( x \) is the independent 
#> variable, and \( b \) is the y-intercept, indicating where the line crosses the
#> y-axis.

options(ellmer_echo = "none") 
cht3 <- chat_openai() 
#> Using model = "gpt-4o".
resp3 <- cht3$chat(prompt)

# Downstream arguments still take precedence
cht4 <- chat_openai(echo = "text") 
#> Using model = "gpt-4o".
resp4 <- cht4$chat(prompt)
#> The equation \( y = mx + b \) represents the equation of a straight line in 
#> slope-intercept form, where \( m \) is the slope of the line and \( b \) is the
#> y-intercept, indicating the point where the line crosses the y-axis.

# Edge cases

# Default to "text" for all args
# This mimics the default behavior of the provider calls
# If not covered, it would fallback to "none"
options(ellmer_echo = c("none", "text", "all")) 
cht5 <- chat_openai() 
#> Using model = "gpt-4o".
resp5 <- cht5$chat(prompt)
#> The equation \( y = mx + b \) represents the equation of a straight line in 
#> slope-intercept form, where \( m \) is the slope and \( b \) is the 
#> y-intercept.

# Correctly throws an error
options(ellmer_echo = c("none", "text")) 
cht6 <- chat_openai()
#> Using model = "gpt-4o".
#> Error in `check_echo()`:
#> ! `echo` must be one of "none", "text", or "all", not "none".
```

<sup>Created on 2025-04-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


hadley:
Perhaps I'm missing something here, but why can't your package call `echo = FALSE` when it calls `$chat`?

dylanpieper:
> Perhaps I'm missing something here, but why can't your package set `echo = FALSE` when it calls `$chat`?

Ah, you're right. This was an oversight on my end. Thanks for the nudge.

PR #429, opened 2025-04-04
Rename `Chat$tokens()` to `Chat$get_tokens()`

### LINKED ISSUE #425: Rename `Chat$tokens()` to `Chat$get_tokens()`
Maybe need to deprecate? Although the data structure is also changing a lot in the dev version so programmatic usage is already broken, and the previous method didn't really give you anything useful so might just be fine to yank.

cc @simonpcouch

hadley:
Fixes #425

@simonpcouch Decided to not bother with deprecation as I think it's rarely used in everyday operation. If it causes other packages to fail, I'll make patches for them.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,9 +52,9 @@
 * `live_browser()` now initializes `shinychat::chat_ui()` with the messages from
   the chat turns, rather than replaying the turns server-side (#381).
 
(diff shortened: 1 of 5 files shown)


simonpcouch (approved):
Rippin🏒

Thanks for the @—will convert rinspect over when this is merged.

PR #438, opened 2025-04-07
Hoist `extra` field up to `Content` class

hadley:
Probably reasonable that every content type has a way to attach arbitrary metadata, and the most obvious way to support Claude's caching. But it's pretty unappealing that I need to adjust the constructor of every subclass in order to preserve the existing behaviour.


hadley:
Alternatively, I could always name the arguments when calling the constructor...

PR #441, opened 2025-04-09
Added custom endpoint for bedrock provider

billsanto:
My university uses a proxy to access AWS with its own domain name. Using provider-azure as an example, I added an "endpoint" parameter to the chat function, which, if non-empty, replaces the standard amazonaws.com URL. Anything that precedes the /model/ segment of the URL would be used for endpoint parameter in chat_aws_bedrock(). We don't include region in the URL, but if others do, then they would want to embed the region into the endpoint argument.

### DIFF
--- a/R/provider-bedrock.R
+++ b/R/provider-bedrock.R
@@ -26,6 +26,11 @@ NULL
 #'   If you're using [cross-region inference](https://aws.amazon.com/blogs/machine-learning/getting-started-with-cross-region-inference-in-amazon-bedrock/),
 #'   you'll need to use the inference profile ID, e.g.
 #'   `model="us.anthropic.claude-3-5-sonnet-20240620-v1:0"`.
+#' @param endpoint Optional. A custom endpoint URL (including protocol,
+#'   hostname, and any base path) to use instead of the default AWS Bedrock
+#'   runtime endpoint constructed from the region (e.g.,
+#'   `https://bedrock-runtime.us-east-1.amazonaws.com`). Trailing slashes
+#'   will be automatically removed.
 #' @param api_args Named list of arbitrary extra arguments appended to the body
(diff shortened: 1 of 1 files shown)


billsanto:
I'd like to add that we now have Bedrock in production and it would be convenient for our R researchers to have Ellmer support, since not everyone has the skills to use Python. Will it be possible to consider this anytime in the near future?

hadley:
Sure, but could you call the argument `base_url` to match all the other chat functions?

R/provider-bedrock.R
```diff
@@ -54,18 +59,24 @@ chat_aws_bedrock <- function(
...
   credentials <- paws_credentials(profile, cache = cache)
-
+  
   model <- set_default(model, "anthropic.claude-3-5-sonnet-20240620-v1:0")
   echo <- check_echo(echo)
 
+  # Clean endpoint URL if provided
+  if (!is.null(endpoint) && nzchar(endpoint)) {
+    endpoint <- sub("/$", "", endpoint)
+  }
```
hadley, line 75: I don't think this is necessary as `req_url_path_append()` will take care of any extra slashes.

R/provider-bedrock.R
```diff
@@ -93,11 +104,13 @@ method(chat_request, ProviderAWSBedrock) <- function(
   tools = list(),
   type = NULL
 ) {
-  req <- request(paste0(
-    "https://bedrock-runtime.",
-    provider@region,
-    ".amazonaws.com"
-  ))
+  # Use custom endpoint if provided, otherwise construct default
+  if (nzchar(provider@base_url)) {
```
hadley, line 108: This default should be constructed in `chat_aws_bedrock()`

hadley:
Hmmm, I've renamed this file since you started working on it, so it might be easier for me just to do this.

PR #458, opened 2025-04-16
Start work on programming vignette

### LINKED ISSUE #290: Document cloning somewhere
Since its such an important pattern to create a conversational tree.  And how to reset existing chats if needed.

hadley:
Also includes a few fixes noticed along the way. Fixes #290

@simonpcouch @dylanpieper you've both probably done the most programming with ellmer so I'd love your feedback 😄


vignettes/programming.Rmd
```diff
@@ -0,0 +1,121 @@
...
+
+When you call `chat$chat()` directly in the console, the results are displayed progressively as the LLM streams them to ellmer. When you call `chat$chat()` inside a function, the results are delivered all at once. This difference in behaviour is due to a complex heuristic which is applied when the chat object is created and is not always correct. So when calling `$chat` in a function, we recommend you control it explicitly with the `echo` argument, setting it to `"none"` if you want no intermediate results to be stream, `"text"` if you want to see what we receive from the assistant, or `"all"` if you want to see both what we send and receive. You likely want `echo = "none"` in most cases:
+
+```{r}
+capital <- function(chat, country) {
+  chat <- chat$clone()$set_turns(list())
+  chat$chat(interpolate("What's the capital of {{country}}"), echo = "none")
+}
+capital(chat, "France")
+```
```
simonpcouch, line 121: May be worth calling out existing integrations of ellmer + streaming into some UI that users can hook into, like shinychat (ellmer + shiny) and streamy (ellmer + rstudioapi).

tests/testthat/test-chat.R
```diff
@@ -80,6 +80,11 @@ test_that("can perform a simple batch chat", {
   expect_equal(chat$last_turn()@contents[[1]]@text, "2")
 })
 
+test_that("can chat with a single prompt", {
+  chat <- chat_openai_test()
+  expect_no_error(chat$chat(interpolate("What's 1 + 1?")))
```
simonpcouch, line 85: 2

simonpcouch (approved):
Will try and remember to leave a note if other things come to mind, but:

* Think of Turns' `@json` field as internal/private; subsetting into it is bad practice and should only be a last resort if you can't derive that information from other fields/methods, as your code likely won't work for a different provider.

dylanpieper:
May be worth mentioning that you can inherit httr2 error classes from the chats using tryCatch. This can be used to provide more useful error information to a user (cross-walked with the API error code description) or do things like create your own retry scaffolding (unless you plan to expose the httr2 retry / transient error list which seems unlikely / unnecessary).

hadley:
@dylanpieper do you have an example of where/why you might do that?

hadley:
@dylanpieper hmmm, if that's a problem, I'd rather make it configurable somehow (maybe an option?)

PR #465, opened 2025-04-17
Use vcr

### LINKED ISSUE #346: Replace current vignette caching approach
They get run too often — need to somehow save ops and replay.

hadley:
Fixes #346

* [ ] Check for other slow examples
* [ ] Pull output handling into separate PR
* [ ] Switch to new vignette style


hadley:
Now done, but in smaller pieces.

PR #471, opened 2025-04-29
Add PortkeyAI support.

maciekbanas:


R/provider-portkey.R
```diff
@@ -0,0 +1,122 @@
...
+  req <- ellmer_req_timeout(req, stream)
+  req <- ellmer_req_user_agent(req)
+  req <- base_request_error(provider, req)
+  req
+}
+
+
+#' @export
+#' @rdname chat_portkey
+models_portkey <- function(
```
shajoezhu, line 95: hi @maciekbanas , i am curious when this get used? i was wondering if you could give an example? Thanks!

shajoezhu (commented):
dear @maciekbanas , thanks you for the implementation, I tested, and it works well! 

let's wait for @hadley 's approval. Thanks!

R/provider-portkey.R
maciekbanas: Hi @shajoezhu this is a separate function/logic from `chat_portkey()` - you can use it to print available models for your PortkeyAI configuration - still as far as I tested it it does not work for temporary accesses in our case (need to get full access to Portkey). This function is based on solutions in other open PR for ellmer: #454

hadley:
@shajoezhu thanks for working on this! Would you mind adding some basic tests, along the lines found in the other providers (they're pretty standardised)? It's not a big deal if you can't get various tests to work (e.g. tool calling can be tricky, and not all models can look at images), you just need to document that those features aren't supported.

maciekbanas:
Hi @hadley - I think that in order to test chat we would need to create a new public account for Portkey app (https://portkey.ai/) and link there for example an OpenAI provider (it's not that hard, but I think it should make use of secrets set up in the project/GHA). We could then use in GHA approach like: `PORTKEY_API_KEY: ${{ secrets.OPENAI_API_KEY }}`

The `ellmer` features for chat will be dependent on model provider(s) we set up and choose. We can document for example that different features are reliant on model provider.

I am also not sure if I should set model by default to `gpt-4o` as for some providers (e.g. bedrock) it won't exist.

hadley:
@maciekbanas yes, that's fine. Just use your own for your tests, and I'll create one for ellmer's. 

You should verify that the major features work at all for Portkey, because my experience with other providers is that they claim things in the docs that aren't actually true.

R/provider-portkey.R
```diff
@@ -11,14 +11,15 @@
...
-#' @param virtual_key Optional. See [documentation](https://portkey.ai/docs/product/ai-gateway/virtual-keys).
+#' @param api_key A Portkey API key.
+#' @param virtual_key A virtual identifier storing LLM provider's API key. See
+#'   [documentation](https://portkey.ai/docs/product/ai-gateway/virtual-keys).
 #' @export
 #' @inheritParams chat_openai
 #' @inherit chat_openai return
 #' @examples
 #' \dontrun{
-#' chat <- chat_portkey()
```
shajoezhu: it worked without virtual key for me actually

shajoezhu: oh I see! Thanks so much for the explanation here!

.github/workflows/live-api.yaml
```diff
@@ -22,6 +22,8 @@ jobs:
       GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
       ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
       AZURE_OPENAI_API_KEY: ${{ secrets.AZURE_OPENAI_API_KEY }}
+      PORTKEY_API_KEY: ${{ secrets.PORTKEY_API_KEY }}
```
hadley, line 25: Actually I realised that I don't actually want to test this on GHA, since the tests are already a bit flaky. I'll just trust that you have run them and they worked 😄

shajoezhu: Thanks a lot @hadley ! I tested it, and it works perfectly!

tests/testthat/test-provider-portkey.R
```diff
@@ -0,0 +1,30 @@
...
+  expect_snapshot(. <- chat_portkey())
+})
+
+test_that("can use images", {
+  chat_fun <- \(...) chat_portkey_test(virtual_key = Sys.getenv("PORTKEY_VIRTUAL_KEY"),
+                                       model = "gpt-4.1-mini", ...)
+
+  test_images_inline(chat_fun)
+  test_images_remote(chat_fun)
+})
```
hadley, line 30: Do you want to test tool calling and structured data extraction too?

.github/workflows/live-api.yaml
maciekbanas: @hadley They worked locally 😎

tests/testthat/test-provider-portkey.R
maciekbanas: Yes, I will add some more tests, just need some time, as I'm tied up with something else at the moment. Thanks!

maciekbanas: Ok, I added tests. Here it is how it looks locally, I get one warning on `async`:

```
> devtools::test_active_file()

══ Testing test-provider-portkey.R ═════════════════════════════════════════════════════════
[ FAIL 0 | WARN 1 | SKIP 0 | PASS 19 ]

── Warning (test-provider-portkey.R:28:3): all tool variations work ────────────
Adding new snapshot:
Code
  chat$chat("Great. Do it again.")
Condition
  Error in `FUN()`:
  ! Can't use async tools with `$chat()` or `$stream()`.
  i Async tools are supported, but you must use `$chat_async()` or `$stream_async()`.

[ FAIL 0 | WARN 1 | SKIP 0 | PASS 19 ]
```

.github/workflows/live-api.yaml
hadley: Perfect, so can you please remove this line?

maciekbanas: Sure! (thought I did it already)

hadley:
Can you please give me permission to edit your PR so I can resolve the merge conflicts?

maciekbanas:
@hadley I sent you invitation to write in the repo. Please check.

hadley:
Sorry I should have explained this more clearly — there's a check box you should see on this PR: 

Screenshot 2025-05-14 at 15 33 42

You just need to check it 😄

maciekbanas:
It seems that this checkbox is not available for me. Maybe because fork is from an organization I am a member of, I'll check it with my colleagues tommorrow.

hadley:
Accepting the invite resolves the problem so I can resolve and merge today 😄

PR #484, opened 2025-05-08
Revise type coercion

hadley:
All optional and absent elements stay as `NULL` except for basic elements inside an array. Based on conversation in #463.


gadenbuie (approved):
> All optional and absent elements stay as NULL except for basic elements inside an array

That makes a lot of sense to me. Looks good!

Issue #143, opened 2024-11-01
Implement batch processing


dylanpieper:
Implement batch processing into ellmer to process lists of inputs through a single prompt. Batch processing enables running multiple inputs to get multiple responses while maintaining safety and ability to leverage ellmer's other features.

**Basic Features** (likely in scope):
- Parallel or sequential API requests
- Prompt caching
- Integration with ellmer's core functionality
  - Tooling and structured data extraction
  - Error, retry, and resume handling
  - Token, rate limit, and throttle handling

**Ambitious Features / Extensions** (likely out of scope):
- Asynchronous batching (polling process)

hadley:
Do you think this code could live in ellmer itself? Do you have any sense of what the UI might look like?

dylanpieper:
After revisiting this, I do think my original ideas were ambitious, and I would like to see this basic functionality in the ellmer package if possible. I revised my original post to simplify the concept.

I think your interface idea is a good start, and as for the returned object, would it be possible to return a list not only of the text answers, but a nested list with each chat object for diagnostics? Although that might be a beefy object if you have big batches (maybe only return the chat object if there was an issue?).

dylanpieper:
@t-emery - Adding Teal because I know he has experience using ellmer for batch processing.

t-emery:
FWIW, my use-case was batched structured data extraction. 

I was reading text data from a tibble and doing structured data extraction (x 18,000).   In my initial attempts, I found that I was using far more tokens than made sense given the inputs.   Eventually I found the workable answer was to clear the turns after each time.  Then everything worked as expected.  

 I'm new to using LLM APIs, so I have great humility that I might be missing something simple.  I haven't had time to think about this deeply yet, and I haven't figured out what part of this is a documentation issue (creating documentation about best practices for running a lot of data) versus a feature issue. 

Here are the relevant functions.   The key fix was:  

`# Clear all prior turns so we start fresh for *each* record
          chat$set_turns(list())`

```r
# 1. Core Classification Function ----
classify_single_project <- function(text, chat, type_spec) {
  result <- purrr::safely(
    ~ chat$extract_data(text, type = type_spec),
    otherwise = NULL,
    quiet = TRUE
  )()
  
  tibble::tibble(
    success = is.null(result$error),
    error_message = if(is.null(result$error)) NA_character_ else as.character(result$error),
    classification = list(result$result)
  )
}

# 2. Process a chunk of data ----
# 3. Modify process_chunk to handle provider types
[...]

PR #496, opened 2025-05-14
First pass at `batch_chat()` implementation

hadley:
Focussing first on Anthropic + OpenAI. Part of #143.

To do:

* [x] Figure out some testing strategy. At least test the flow with some mocked methods?
* [x] Implement `batch_chat_structured()`
* [x] Think about if we want to store a hash of the chat + prompts to warn the user if there's a mismatch
* [x] Better progress bar

@gadenbuie While the implementation isn't 100% complete (and I haven't written any tests yet), I'd love to get your thoughts on this API. In particular, what do you think of the required `path` argument, and is it correct to throw an error if `wait = FALSE`.


R/batch-chat.R
```diff
@@ -0,0 +1,129 @@
...
+    state$batch <- batch_poll(provider, state$batch)
+    status <- batch_status(provider, state$batch)
+    save_state()
+
+    if (status$working && wait) {
+      cli::cli_progress_bar(
+        format = paste(
+          "{cli::pb_spin} Processing... ",
+          "{status$n_processing} -> {cli::col_green({status$n_succeeded})} / {cli::col_red({status$n_failed})} ",
+          "[{cli::pb_elapsed}]"
```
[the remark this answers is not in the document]
gadenbuie: Another minor comment: I'd prefer using words in addition to colors to communicate status here and I think there's room for it.

```
# Instead of
⠙ Processing...  46 -> 4 / 0 [5.6s]

# Something like this
⠙ Processing...  46 pending | 4 done | 0 failed [0:05]
```

hadley:
The problem is that you can't in store the provider on disk because it's includes credentials. And if we only stored part of the provider we'd need some relatively complicated code to stitch it back together on reload; that feels fragile to me. 

We could _maybe_ serialise the remained of the chat object to json, but that's going to be quite a bit of work, and isn't super compelling if you still have to pass in the `provider`.

R/batch-chat.R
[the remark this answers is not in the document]
hadley: Oh yeah, good idea. Let me figure out how to do that.

hadley:
BTW that is an awesome example!

gadenbuie:
> The problem is that you can't in store the provider on disk because it's includes credentials. And if we only stored part of the provider we'd need some relatively complicated code to stitch it back together on reload; that feels fragile to me.

What if it's more of a best-effort approach? Say we just store the provider name, e.g. `"provider": "openai"` or `"anthropic"` and then we call the default chat function based on that value.

If that doesn't work, `batch_chat_status()` and `batch_chat_results()` would take an optional `chat` argument where users could pass in the appropriate chat object.

hadley:
But that would mean that you lose the model information. Obviously, you could of course track that too, but then you're on a slippery slope. Do you think having the user re-create the chat object is really that painful?

gadenbuie:
> But that would mean that you lose the model information.

I think we _could_ track the model information (and maybe `params`, too, if those don't carry credentials). Or we might not need to track the model – that's included in the model results (at least for OpenAI and Anthropic, but I'd imagine most would do that).

> Do you think having the user re-create the chat object is really that painful? ... you can just re-run exactly what you ran last time

I'm not thinking it's _that hard_ to recreate the `chat` object, but checking the status and getting the results only requires the provider to speak with the API. I do think we should consider it hard to recreate the prompts since batching is designed to scale up to 100k prompts. Which speaks to also including `chat$get_turns()` and `prompts` in the serialized state.

> but then you're on a slippery slope.

I can certainly see the slope and that it's slippery, but I do think it's worth serializing and storing what we can.

I guess what I'm saying is that I think `batch_chat()` should prioritize two use cases:

1. Interactive batch completions, primarily for cost reduction.
2. Entirely async batch completions that might scale to 10k completions.

In the second case, I'm thinking about the fact that your batch could take up to 24 hours to complete (likely less, but possible at scale) and that there could be days between submitting the request and retrieving the result (Anthropic gives you ~30 days to retrieve results).

Also, in the second case at scale, it seems much less likely that you're going to continue the conversation on any of the chat results, so it seems okay to me that the returned list of chats serve as a container for the completion. And if it's important to carry on the conversation you can always recreate the base chat object.

hadley:
@t-kalinowski @gadenbuie thanks for the feedback! Given the constraint that I am submitting ellmer to CRAN and I don't believe that this is going to be a highly used feature, I'm going to stick to the current shape of the API. I don't have time to figure out how to safely serialize and deserialize a chat/provider, and this has to be done carefully to avoid leaking credentials (and not create a maintenance headache going forward). I'm also not convinced that generating even 100k prompts is typically going to be that expensive, particularly since you will presumably want to align the LLM generated outputs with your inputs. So I think we just need to get this into the hands of users and see where their pain points are.

I will save a hash of the chat and prompts so that we can avoid the most obvious footguns. 

I do think it will be useful to have some kind of per-chat caching, but that's out of scope for ellmer, since you're going to have to much a bunch of design decisions specific for your use case.

gadenbuie:
I started writing this probably at the same time as your last response. I totally agree with your conclusion, but thought I'd add this for future work/consideration

I just talked with @schloerke, and I've realized that the slope is indeed quite slippery but some of that Shiny will have to navigate to enable bookmarking existing chats, which mostly comes down to serializing `chat$get_turns()`. (And in the Shiny contexts, the user will still need to pass in a bare `chat_*()` object.) So in the long run, I'd advocate for additionally storing

1. provider name
2. model name
3. `chat$get_system_prompt()`
4. `prompts` (just text for now, maybe richer with Barret's work)
5. `chat$get_turns()` (not now, but when Barret's work is ready)

Automatically creating the `chat` object feels like a nice affordance, but after reflecting I think that requiring `chat` is a solid choice. The API could even be simplified to:

```r
# Start a batch request
batch_chat(chat, prompts, wait = FALSE)
#> Batch started. State in "ellmer-batch-abc123.json"

# Check on that batch request if you still have `prompts`
batch_chat(chat, prompts)

# Or check on it later without prompts
batch_chat(chat, path = "ellmer-batch-abc123.json")
⠙ Processing... [0:05] 46 pending | 4 done | 0 failed 

# Or get the results without prompts
chats <- batch_chat(chat, path = "ellmer-batch-abc123.json")
```

hadley:
@gadenbuie that reminded me to file https://github.com/tidyverse/ellmer/issues/499.

And thanks for that API sketch — I think that confirms that the current API won't be unnecessarily limiting if we save more state to disk in the future.

schloerke:
Related "to JSON" record / replay PR: https://github.com/tidyverse/ellmer/pull/503/files#diff-cdd4c06a829b36da7fef0711b5523715d58b3a33abcb29c861cc95d5a8a9c1ff

R/batch-chat.R
```diff
@@ -0,0 +1,333 @@
...
+#' data <- batch_chat_structured(
+#'   chat = chat,
+#'   prompts = prompts,
+#'   path = "people-data.json",
+#'   type = type_person
+#' )
+#' data
+#' }
+#'
+batch_chat <- function(chat, prompts, path, wait = TRUE) {
```
gadenbuie, line 68:
```suggestion
#'
#' @export
batch_chat <- function(chat, prompts, path, wait = TRUE) {
```

R/batch-chat.R
```diff
@@ -0,0 +1,333 @@
...
+#' you're working with a different provider, you may want to use
+#' [parallel_chat()] instead.
+#'
+#' Since batched requests can take a long time to complete, `batch_chat()`
+#' requires a file path that is used to store information about the batch so
+#' you never lose any work. You can either set `wait = FALSE` or simply
+#' interrupt the waiting process, then later, either call `batch_chat()` to
+#' resume where you left off or call `batch_chat_completed()` to see if the
+#' results are ready to retrieve. Once you have the results, you can delete
+#' the file.
```
gadenbuie, line 21: Maybe mention that you could also keep the file to cache the results?

And maybe also that `batch_chat()` stores the results in the file (rather than retrieving them anew from the provider) so as long as you have the file you can access the results. Re-running a script and using cached `batch_chat()` results each time as Tomasz mentioned sounds both like a feature and a footgun.

R/batch-chat.R
```diff
@@ -0,0 +1,333 @@
...
+    job$stage,
+    "submitting" = FALSE,
+    "waiting" = !job$poll()$working,
+    "retrieving" = TRUE,
+    "done" = TRUE,
+    cli::cli_abort("Unexpected stage: {job$stage}", .internal = TRUE)
+  )
+}
+
+BatchJob <- R6::R6Class(
```
gadenbuie, line 144: Really like the R6 approach!

R/batch-chat.R
```diff
@@ -0,0 +1,333 @@
...
+    wait = function() {
+      # always poll once, even when wait = FALSE
+      status <- self$poll()
+
+      if (self$should_wait) {
+        cli::cli_progress_bar(
+          format = paste(
+            "{cli::pb_spin} Processing... ",
+            "{status$n_processing} -> {cli::col_green({status$n_succeeded})} / {cli::col_red({status$n_failed})} ",
+            "[{self$elapsed()}]"
```
gadenbuie, line 253:
```suggestion
            "{cli::pb_spin} Processing...",
            "[{self$elapsed()}]",
            "{status$n_processing} pending |", 
            "{cli::col_green({status$n_succeeded})} done |",
            "{cli::col_red({status$n_failed})} failed"
```

image

As a user, I think this format is much friendlier because I never have to guess about what the numbers mean. I do think they're _eventually_ clear, but it's nice to just not have to ask the question in the first place.

R/batch-chat.R
```diff
@@ -0,0 +1,333 @@
...
+          )
+        )
+        while (status$working) {
+          Sys.sleep(0.5)
+          cli::cli_progress_update()
+          status <- self$poll()
+        }
+        cli::cli_process_done()
```
gadenbuie, line 261:
```suggestion
        cli::cli_progress_done()
```

R/batch-chat.R
hadley: Doh

capitals.json
gadenbuie, line 1: I don't think you intended to commit this

tests/testthat/_snaps/batch-chat.md
```diff
@@ -0,0 +1,37 @@
+# errors if chat/provider/prompts don't match previous run
+
+    Code
+      batch_chat(chat, prompts, path)
+    Condition
+      Error in `batch_chat()`:
+      ! provider, prompts, and user_turns don't match stored values.
```
gadenbuie, line 7: Should the error message include the path value to make it easier to track down the problematic file?

tests/testthat/_snaps/batch-chat.md
```diff
@@ -0,0 +1,37 @@
...
+    Code
+      batch_chat(chat_ollama)
+    Condition
+      Error in `batch_chat()`:
+      ! Batch requests are not currently supported by this provider.
+    Code
+      batch_chat(chat_openai, "a")
+    Condition
+      Error in `batch_chat()`:
+      ! `prompts` must be a list or prompt, not the string "a".
```
gadenbuie, line 26: Is "list or prompt" expected in the error message? Or is it supposed to be "list of prompts"?

tests/testthat/test-batch-chat.R
```diff
@@ -0,0 +1,125 @@
...
+  prompts <- list(
+    "What's the capital of Iowa?",
+    "What's the capital of New York?",
+    "What's the capital of California?",
+    "What's the capital of Texas?"
+  )
+  chats <- batch_chat(
+    chat,
+    prompts,
+    path = test_path("batch/state-capitals.json")
```
gadenbuie, line 13: Should this file be committed? Or cleared out before this test?

tests/testthat/test-batch-chat.R
```diff
@@ -0,0 +1,125 @@
...
+    path = test_path("batch/state-name.json"),
+    type = type_state
+  )
+  expect_equal(nrow(data), 4)
+})
+
+test_that("errors if chat/provider/prompts don't match previous run", {
+  chat <- chat_anthropic_test(system_prompt = "Be cool")
+  prompts <- list("What's the capital of Iowa?")
+  path <- test_path("batch/state-capitals.json")
```
gadenbuie, line 30: Instead of using the file from the previous test, you could store a small JSON with enough state information to trigger the hash checks.

gadenbuie (approved):
Looks great! I really like the feature and the internal API with R6 class is quite clean. No big comments, all just about final polish

capitals.json
hadley: Oops

tests/testthat/_snaps/batch-chat.md
hadley: Hmmm, the condition is `!is.list(prompts) && !is_prompt(prompts)` so the message is technically correct albeit slightly confusing.

hadley:
@gadenbuie thanks for the thoughtful feedback, it's appreciated!

bshor:
batch_chat works great for OpenAI models (haven't tried Claude). But I noticed that Google has enabled batch, not sure when. Here's [an example](https://ai.google.dev/gemini-api/docs/pricing) from it's documentation.

When you have resources, it would be great if batch_chat could be extended to Gemini models, given how well the latter are performing. No pressure, of course!

hadley:
@bshor can you please file an issue?

PR #504, opened 2025-05-15
Use `map_dbl` shim instead of `map_vec()`

hadley:
And re-format with air.

@maciekbanas can you please double check that this is correct? ellmer does not import purrr; it just uses a small set of wrappers around base R functions.


maciekbanas:
@hadley just checked. Works perfectly!

PR #513, opened 2025-05-16
add support for `GEMINI_API_KEY`

t-kalinowski:


R/provider-gemini.R
```diff
@@ -562,16 +565,31 @@ merge_gemini_chunks <- merge_objects(
...
+  }
+
+  if (gemini && api_key == "") {
+    api_key <- c(gemini = Sys.getenv("GEMINI_API_KEY"))
+  }
+
+  if (nzchar(api_key)) {
+    if (identical(names(api_key), "gemini")) {
+      creds <- list(key = unname(api_key))
+      attr(creds, "auth_location") <- "query"
```
hadley, line 586: Can you tell me more about this?

t-kalinowski: Based on the quickstart example, it looks like the only way to pass the `GEMINI_API_KEY` is as a query param, not in the headers. In ellmer, I don't see a way to do that right now, so I used an attr for this edge case.

https://ai.google.dev/gemini-api/docs/quickstart?lang=rest

I'll take a look through `google.genai` to see if there's a way to pass the Gemini API key via headers.

t-kalinowski: 🤦🏻 Seems like we just pass the gemini api key the same way as the google key

PR #514, opened 2025-05-16
Reformat with dev Air, embrace the autobrace

DavisVaughan:
Testing out Air's new autobracing capability

This is all backwards compatible with the current version of Air, so even if you format again with Air 0.5.0 (release air), then nothing will change, so it is safe to merge.

Nice to see these little code improvements! They all look like real improvements to me.

### DIFF
--- a/R/provider-openrouter.R
+++ b/R/provider-openrouter.R
@@ -116,19 +116,18 @@ check_openrouter_error <- function(error, call = caller_env()) {
 }
 
 method(chat_resp_stream, ProviderOpenRouter) <- function(provider, resp) {
-  repeat
-    ({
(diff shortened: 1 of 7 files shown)


R/provider-openrouter.R
```diff
@@ -116,19 +116,18 @@ check_openrouter_error <- function(error, call = caller_env()) {
 }
 
 method(chat_resp_stream, ProviderOpenRouter) <- function(provider, resp) {
-  repeat
-    ({
```
DavisVaughan, line 120: Idk what was going on with the `({` here, but I remove the `(` and it reflowed properly

hadley:
LGTM!

PR #526, opened 2025-05-22
Add Open Telemetry instrumentation

atheriel:
This commit wraps all LLM model calls in an Open Telemetry span that abides by the (still nascent) [semantic conventions for Generative AI clients][0].

It's very similar in approach to what was done for `httr2`, and in fact the two of them complement one another nicely:
https://github.com/r-lib/httr2/pull/729.

For example:

    library(otelsdk)

    Sys.setenv(OTEL_TRACES_EXPORTER = "stderr")

    chat <- ellmer::chat_databricks(model = "databricks-claude-3-7-sonnet")
    chat$chat("Tell me a joke in the form of an SQL query.")

[0]: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/


atheriel:
Traces that mix LLM model call spans with `httr2` spans:

Screenshot 2025-05-22 at 14-25-20 Live · posit1_starter-project · Pydantic Logfire

jcheng5:
cc @cpsievert @schloerke @icarusz

hadley:
Do we want to (optionally?) also include user and assistant messages, a la https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/ ?

atheriel:
@hadley I do. But there's a ton of disagreement in the OTel LLM community about how to do that, and none of the existing instrumentation libraries work in the same way :disappointed:. Plus the whole "structured body" mechanism the current spec proposes (1) isn't supported by the span API; and (2) is formally deprecated.

So I kind of think we need to noodle on what to do there, and I suggest pushing it into follow-up work. I'm planning on writing up an issue describing what options we have.

I also think we should have first-class support for [tool call spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/#execute-tool-span), because that's something that `ellmer` focuses on specifically. This PR is really the "basic" bit that the title implies.

atheriel:
Moving this back to draft because it has known issues (i.e. the concurrency does not work correctly).

atheriel:
This has been updated to support async operations and for changes in `otel` and `otelsdk`. It now also includes pretty extensive unit tests and support for agent and tool call spans.

R/otel.R
```diff
@@ -0,0 +1,159 @@
...
+
+# Inspired by httr2:::get_tracer() / shiny:::get_tracer()
+# Using local scope avoids an environment object lookup on each call.
+default_tracer <- local({
+  tracer <- NULL
+  function() {
+    if (!is.null(tracer)) {
+      return(tracer)
+    }
+    if (testthat::is_testing()) {
```
schloerke: Must check if `{testthat}` is installed (as it is currently only a `Suggests` pkg)

hadley: You can just inline that function (i.e. copy it in into `utils.R`)

hadley:
One other question: do we want to log something about auth here (i.e. in particular, which auth was automatically picked?)

R/otel.R
```diff
@@ -0,0 +1,159 @@
...
+      promises::local_ospan_promise_domain(activation_scope)
+    }
+    otel::local_active_span(
+      ospan,
+      end_on_exit = FALSE,
+      activation_scope = activation_scope
+    )
+    # For some reason, when using `end_on_exit = TRUE` above, an error would occur during `spn$end(status = "auto")`. When using `withr::defer()` here, it works fine.
+    # TODO: Set status?
+    withr::defer(promises::end_ospan(ospan), envir = activation_scope)
```
shikokuchuo: Suggest we use the leaner standalone withr defer (as we do in promises).

R/otel.R
```diff
@@ -0,0 +1,159 @@
...
+    promises::create_ospan(
+      sprintf("chat %s", provider@model),
+      tracer = tracer,
+      options = list(
+        parent = parent_ospan,
+        kind = "CLIENT"
+      ),
+      attributes = list(
+        "gen_ai.operation.name" = "chat",
+        "gen_ai.system" = tolower(provider@name),
```
shikokuchuo:
```suggestion
        "gen_ai.provider.name" = tolower(provider@name),
```
As per the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/). This is evolving fast so it might well be a recent change.

R/otel.R
```diff
@@ -0,0 +1,159 @@
...
+  local_envir = parent.frame()
+) {
+  agent_ospan <-
+    promises::create_ospan(
+      "invoke_agent",
+      tracer = tracer,
+      options = list(kind = "CLIENT"),
+      attributes = list(
+        "gen_ai.operation.name" = "chat",
+        "gen_ai.system" = tolower(provider@name)
```
shikokuchuo:
```suggestion
        "gen_ai.provider.name" = tolower(provider@name)
```

DESCRIPTION
```diff
@@ -30,7 +30,8 @@ Imports:
     jsonlite,
     later (>= 1.4.0),
     lifecycle,
-    promises (>= 1.3.1),
+    otel (>= 0.2.0),
+    promises (>= 1.3.3.9004),
```
shikokuchuo:
```suggestion
    promises (>= 1.5.0),
```
🥳

DESCRIPTION
```diff
@@ -52,13 +53,16 @@ Suggests:
     withr
 VignetteBuilder:
     knitr
+Remotes:
+    rstudio/promises
```
shikokuchuo:
```suggestion
```

R/httr2.R
```diff
@@ -23,18 +26,35 @@ chat_perform <- function(
...
+    parent_otel_span = NULL
+  ) {
+    setup_active_promise_otel_span(parent_otel_span)
+
     resp <- req_perform_connection(req)
     on.exit(close(resp))
 
     repeat {
+      # Ensure the span is active for each await/yield point
+      setup_active_promise_otel_span(parent_otel_span)
```
schloerke: Q: Is this necessary?

R/httr2.R
```diff
@@ -47,11 +67,20 @@ on_load(
...
+  ) {
+    setup_active_promise_otel_span(parent_otel_span)
+
     resp <- req_perform_connection(req, blocking = FALSE)
     on.exit(close(resp))
 
     repeat {
+      # Ensure the span is active for each await/yield point
+      setup_active_promise_otel_span(parent_otel_span)
+
```
schloerke: Q: Is this necessary?

The goal is to make sure httr2 has a parent span and I believe that is done within `req_perform_connection()`.

I don't believe we need to reactivate to help with streaming the result back. 

The only important part is request... not the stream.

R/otel.R
schloerke: Also applied to `local_tempfile()`

R/otel.R
schloerke: Already there. `is_testing()`

R/chat-tools.R
```diff
@@ -33,7 +33,8 @@ on_load({
     echo = "none",
     on_tool_request = function(request) invisible(),
     on_tool_result = function(result) invisible(),
-    yield_request = FALSE
+    yield_request = FALSE,
+    parent_otel_span = NULL
```
schloerke: Rename these variables to `otel_span` for better less confusion. The name is too specific

shikokuchuo:
Just adding a comment that I've adopted @schloerke's suggestions from #848 (with a couple of extensions as I noted) in [8080dc6](https://github.com/tidyverse/ellmer/pull/526/commits/8080dc6617a1e8f183f8b70931f836547deb2247), and this is a common approach to fixing the tracer across the packages we're instrumenting.

R/httr2.R
shikokuchuo: For the non-async streaming case, there are no promises used in the generator anyway.

R/httr2.R
shikokuchuo: Indeed, even in the async streaming case, these are 'internal' and do not need the promise domain to be active. I've removed these lines in [01f20e6](https://github.com/tidyverse/ellmer/pull/526/commits/01f20e6108aad0e110e8200f89db7f14f250378f).

shikokuchuo:
@schloerke one of the live-api [test failures](https://github.com/tidyverse/ellmer/actions/runs/19197533061/job/54880658925?pr=526#step:6:224) concerns a test on this block:
```r
p1 <- chat$chat_async("What's the current date in Y-M-D format?") |>
  promises::then(function(result) {
    chat$chat_async("What date will it be 47 days from now?")
  })
```
The test expectation is for both agent spans to be top level.

I'm thinking this was set up pre-otel promise domains, and want to confirm with you that one should indeed be a child span of the other given it executes within a `$then()` context. Thanks!

schloerke:
> The test expectation is for both agent spans to be top level.
> 
> I'm thinking this was set up pre-otel promise domains, and want to confirm with you that one should indeed be a child span of the other given it executes within a `$then()` context. Thanks!

I'd still expect both `chat$chat_async()` calls to be top level.

**Investigating

schloerke:
The error was introduced in [`eeb0b84` (#526)](https://github.com/tidyverse/ellmer/pull/526/commits/eeb0b8408c3be85def91a528bbfebbbfc1c0e0fd#diff-0cc4a76032b57fcc125d41dfa3fb0f0c39976bb00a1d84bb56a0b77c331ce2d1R166) 🥸 

But!.. It's fixed now. Thank goodness for unit tests

---------------------

Restrictions on the other earlier failing test have been relaxed and now pass

shikokuchuo (approved):
Thanks @schloerke. I've made a couple more tidy-ups, the only consequential one to move otel to 'suggests', consistent with the other packages we're instrumenting.

Apart from the unit tests themsleves, I've also been testing using your shinychat demo (Shiny, ellmer, httr2 and mirai spans)  and the traces all look good.

shikokuchuo:
@hadley this PR is now ready for review. Thanks!

schloerke:
**Friendly poke to get this PR merged

R/chat-tools.R
```diff
@@ -155,19 +157,25 @@ invoke_tool <- function(request) {
     return(args)
   }
 
+  tool_span <- local_tool_otel_span(
```
hadley: Can this (and below) fit on one line?

hadley:
Could we get a longish news bullet that describes what gets logged? And can someone check that https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/ hasn't changed in the almost one year (!!) that this PR has been open?

shikokuchuo:
[d618297](https://github.com/tidyverse/ellmer/pull/526/commits/d618297d48a918c02d987a33b70742f1b11a6b68) adds input and output tokens as per the semantic conventions.

I had to move Anthropic's 1.25x multiplier out so it's just used for value calcs, and doesn't distort the actual token counts. Added a regression test for this.

Screenshot 2026-05-01 at 10 38 20

shikokuchuo:
[c9e4e0e](https://github.com/tidyverse/ellmer/pull/526/commits/c9e4e0ec260f012507047b26b6bd836ee41fa62e) gives us: `gen_ai.input.messages`, `gen_ai.output.messages`, and `gen_ai.system_instructions`.

The semantic conventions require this to be opt-in (via the `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` env var) since these payloads may contain user data.

These unlock a special gen ai UI in Logfire so we get something nice like this:

Screenshot 2026-05-01 at 11 57 28


I've noticed a few rough edges where occasionally the UI will fall back to raw json output, but I don't think it's significant enough to stop us shipping an initial version of this functionality.

Issue #527, opened 2025-05-23
Allow JSON string to structured chat `type`


hafen:
I work in an organization where non-R developers have defined JSON schemas that we want to use to prescribe the output from structured chats in some pipelines we are building with ellmer.

These schemas can be pretty complex and also change over time and it would be quite cumbersome to continually manually translate these schemas to the `type_*()` definitions in R (which then in turn get translated back to JSON anyway).

To address this use case, could ellmer allow a JSON string type as an argument to `type` in `$structured_chat()`? If this seems acceptable I'd be happy to work on a PR.

PR #529, opened 2025-05-27
Add support for type = json_schema() in structured chat - 

### LINKED ISSUE #133: Provide way to use pre-existing JSON schema
e.g. https://docs.posit.co/connect/api/swagger.json

hafen:
Closes #133 (see also #527)

This is the most minimal and least intrusive solution I could come up with.

Perhaps the only downside is that `TypeJsonSchema` inherits from `Type` while it really isn't a type which could be confusing. But doing it this way makes it convenient for working with things like `wrap_type_if_needed()` without issue and for deferring dealing with the content until its `as_json()` method is called. I did not prepend the wrapper function `json_schema()` with `type_` to help avoid this confusion with the user.

I'm submitting in this state to make sure this is an acceptable approach. If so, I can update tests, news, pkgdown yaml, and docs accordingly.

If this is a good approach, I think the only question I have is whether `json_schema()` belongs in `@rdname type_boolean` with the others or not.

Also I don't have all providers set up locally so a lot of provider-specific tests were skipped on my local test runs.

Here's an example:

``` r
library(ellmer)

chat <- chat_openai()
#> Using model = "gpt-4.1".

# with types
chat$chat_structured(
  "My name is Susan and I'm 13 years old",
  type = type_object(
    name = type_string(),
    age = type_number()
  )
)
#> $name
#> [1] "Susan"
#> 
#> $age
#> [1] 13

# with json schema (text, file path,  or url)
chat$chat_structured(
  "My name is Susan and I'm 13 years old",
  type = json_schema('
{
  "type": "object",
  "description": "",
  "properties": {
    "name": {
      "type": "string",
      "description": ""
    },
    "age": {
      "type": "number",
      "description": ""
    }
  },
  "required": [
    "name",
    "age"
  ],
  "additionalProperties": false
}'
  )
)
#> $name
#> [1] "Susan"
#> 
#> $age
#> [1] 13
```

<sup>Created on 2025-05-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/as-json.R
```diff
@@ -36,3 +36,7 @@ method(as_json, list(Provider, TypeArray)) <- function(provider, x) {
     items = as_json(provider, x@items)
   )
 }
+
+method(as_json, list(Provider, TypeJsonSchema)) <- function(provider, x) {
```
hadley, line 40: We'll need to do a bit more if we also want this to work with `tool`/`ToolDef` (which I think we do for the sake of consistency). shis will need a little extra work in order if we also want to allow json schema in tool specs. That will require some work on the UI side at `tool()` (since currently you supply named arguments) and for the `as_json(*, ToolDef)` methods.

hadley (approved):
I think I'm convinced by your argument that it's less important for tool calls, since there has to be an R function involved, so it's less likely to have an external schema. 

So I think that means your PR is complete, and just needs a bullet at the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hafen:
> I think I'm convinced by your argument that it's less important for tool calls, since there has to be an R function involved, so it's less likely to have an external schema.
> 
> So I think that means your PR is complete, and just needs a bullet at the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Great! Just updated news.

hadley:
Thanks!

SokolovAnatoliy:
hi @hafen  and @hadley. Thanks for a great feature. I wanted to report a small bug when using this feature with a deeply nested JSON schema. 

It seems that there is an openAI schema limitation that does not allow for object nesting deeper than 5 levels. If you use more levels than that, you get an error like this:
```
ℹ : Invalid schema for response_format 'structured_data': 6 levels of nesting exceeds limit of 5.
```
However, what I found is that the current implementation of the `chat_structured` function contains the following lines:

```
needs_wrapper <- S7_inherits(private$provider, ProviderOpenAI)
type <- wrap_type_if_needed(type, needs_wrapper)
```
The result is that the passed schema is wrapped one more time as an object. This results in one more additional nesting level, that should not be there.

I confirmed that if I manually set  `needs_wrapper <- FALSE`, then the `structured_chat` call works fine.


Here is a reprex that you can use as a test for this behavior.

```
chat <- chat_openai(model = "o3",  system_prompt = "You are a baker")

recipe <- "Preheat the oven to 350 degrees F (175 degrees C)."

result <- chat$chat_structured(recipe,
                               type = ellmer::type_from_schema('{
  "title": "Cookie Schema",
  "description": "Test",
  "type": "object",
  "required":["recipes"],
  "additionalProperties": false,
  "properties":{
    "recipes": {
      "type": "array",
      "description": "List of detailed information on each of recipes",
      "items":{
        "type": "object",
        "description": "Detailed information about a specific recipe.",
        "required": [ "equipment"],
        "additionalProperties": false,
        "properties": {
          "equipment": {
            "type": "array",
            "description": "Detailed list of all equipment required.",
            "items": {
              "type": "object",
              "properties": {
                "details": {
                  "type":"object",
                  "description": "Specifications and operating conditions.",
                  "required": ["specification"],
                  "additionalProperties": false,
                  "properties": {
                    "specification": {
                      "type": "object",
                      "description": "Equipment specifications like temperature range",
                      "required": ["temperature"],
                      "additionalProperties": false,
                      "properties": {
                        "temperature": {"type": "string", "description": "e.g., 100F, 500F"}
                      }
                    }
                  }
                }
              },
              "additionalProperties": false,
              "required": ["details"]
            }
          }
        }
      }
    }
  }
}
'))
```

This is the expected result (when setting the `needs_wrapper <- FALSE` inside the `chat_structured` function.

```
$recipes
$recipes[[1]]
$recipes[[1]]$equipment
$recipes[[1]]$equipment[[1]]
$recipes[[1]]$equipment[[1]]$details
$recipes[[1]]$equipment[[1]]$details$specification
$recipes[[1]]$equipment[[1]]$details$specification$temperature
[1] "350F"
```

hadley:
@SokolovAnatoliy can you please file an issue?

Issue #520, opened 2025-05-21
Typo in example for google_upload


gbiele:
First, thanks for working at this package!

There is a minor typo in the example for *google_upload*, where the file is uploaded to gemini/google, but then the chat used to ask questions about the pdf is an *openai* chat.

Here is a quote from the file:

```r
## Not run: 
file <- google_upload("path/to/file.pdf")

chat <- chat_openai()
chat$chat(file, "Give me a three paragraph summary of this PDF")

## End(Not run)
```

Best, Guido

hadley:
Oops. Do you want to have a go at a PR to fix this?

hadley:
Actually looks like it's already done in #531

PR #531, opened 2025-05-28
Switch to chat_google_gemin() in example

bgreenwell:
The upload to Google Gemini example is followed by connecting to the OpenAI provider, which seems incorrect. Switching to connect to Google Gemini via the associated chat constructor.


hadley:
Fixes #520

hadley:
Thanks @bgreenwell!

PR #539, opened 2025-05-29
Default to the latest Claude model in chat_snowflake()

atheriel:
This aligns to what we're doing in `chat_anthropic()`.


atheriel:
Unfortunately Snowflake uses a slightly different naming convention that is incompatible with Anthropic :roll_eyes:.

It's also a bit of an issue that Snowflake [rolls out models](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-llm-rest-api#model-availability) at a slightly different pace than Anthropic, and I think it's wise to default to a model available in most Snowflake regions (which is not always the case, and the main reason why we didn't default to Claude 3.5 Sonnet before). They've also got this complex cross-region thing to consider. All in all I think we'll want to have a slightly different cadence here, at least for now.

hadley (approved):
Of course they do 🙄 But good point about the roll out speed.

PR #549, opened 2025-05-29
Fix tool calling for chat_databricks()

### LINKED ISSUE #548: `chat_databricks()` barfs when tool calls are registered
It seems that Databricks's tool serialization is not OpenAI compatible after all:

```r
> chat <- chat_databricks(model = "databricks-claude-3-7-sonnet")
> chat$register_tool(tool(
  function() format(Sys.Date()),
  .name = "current_date",
  .description = "Returns the current date in ISO 8601 format."
))
> chat$chat("What's the current date in Y-M-D format?")
Error in `req_perform_connection()` at elmer/R/httr2.R:37:5:
! HTTP 400 Bad Request.
• tools.0.custom.strict: Extra inputs are not permitted
```

Reported by @blairj09.

atheriel:
This commit fixes up the tool definitions we send to Databricks so that they actually work.

Also: despite their documentation, multi-step tool calling *does* seem to work now. So this is a significant upgrade in terms of possibilities with this provider.

Unit tests for tool calling have been re-enabled and there are some new ones to ensure we get the format right.

Closes #548.

### DIFF
--- a/R/provider-databricks.R
+++ b/R/provider-databricks.R
@@ -23,8 +23,7 @@
 #' ## Known limitations
 #'
 #' Databricks models do not support images, but they do support structured
-#' outputs. Tool calling support is also very limited at present and is
-#' currently not supported by ellmer.
+#' outputs and tool calls for most models.
 #'
 #' @family chatbots
 #' @param workspace The URL of a Databricks workspace, e.g.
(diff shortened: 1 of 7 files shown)


R/provider-databricks.R
```diff
@@ -177,6 +176,42 @@ method(as_json, list(ProviderDatabricks, ContentText)) <- function(
...
+  # what their documentation says. It *is* supported for structured outputs,
+  # though. I suspect a copy & paste error in their docs.
+  compact(list(
+    type = "function",
+    "function" = compact(list(
+      name = x@name,
+      description = x@description,
+      # Use the same parameter encoding as the OpenAI provider, but only if
+      # there actually are parameters.
+      parameters = if (length(x@arguments@properties) != 0)
```
hadley, line 191: I'm suprised that you need the `if` here — I would have though that the output of `as_json()` would be `NULL` and thus dropped by the `compact()` on line 184.

hadley (approved):

R/provider-databricks.R
atheriel: No, the output of `as_json()` is a kind of weird, empty object:

```r
chat <- chat_databricks()
tool <- tool(
   function() format(Sys.Date()),
   .name = "current_date",
   .description = "Returns the current date in ISO 8601 format."
)
as_json(chat$get_provider(), tool@arguments) |> jsonlite::toJSON(auto_unbox = TRUE)
```
```
{"type":"object","description":"","properties":{},"required":[],"additionalProperties":false} 
```

I didn't want to mess with that for fear of breaking one of the other OpenAI providers.

hadley: Interesting.

PR #551, opened 2025-05-29
Redact API keys when saving to disk

### LINKED ISSUE #534: Too easy to save chat object, accidentally exposing API key
Try using https://github.com/r-lib/httr2/issues/721#issuecomment-2851426676

hadley:
Using the magic of weak refs. Fixes #534


simonpcouch (approved):
SO GOOD YAY

Just to make sure my mental model is right here—this doesn't cause any hiccups with sending chat apps into production, right? e.g. in [Joe's ellmer assistant](https://jcheng.shinyapps.io/ellmer-assistant/), the weak reference is created in the server each time the app starts up?

atheriel (approved):
Nice!

PR #558, opened 2025-06-04
Fix a `chat_snowflake()`/`chat_databricks()` mixup in the 0.2.1 NEWS

atheriel:
Tool calling support landed for `chat_databricks()`, not `chat_snowflake()`.


hadley (approved):
Oops, I'm sure that was my mistake.

Issue #536, opened 2025-05-28
Update documentation for `echo` argument


howardbaik:
I’ve been looking through the ellmer source code and noticed that the [`echo` argument](https://github.com/tidyverse/ellmer/blob/main/R/provider-openai.R#L26-L30) no longer supports ["text"](https://github.com/tidyverse/ellmer/blob/2c8364f5efddca492334a2ea9b68ed70cea06f29/R/utils.R#L64-L71) as a valid option. Is there a reason "text" is still mentioned in the function documentation? If not, I'd be happy to submit a PR to update it to `output` instead.

howardbaik:
@gadenbuie You beat me to it! Thanks for letting me know.

PR #565, opened 2025-06-11
Change `text` to `output`

howardbaik:
This is an update to the "Programming with ellmer" vignette to reflect the change in echo argument of `text` to `output`. 

Related to https://github.com/tidyverse/ellmer/issues/536.


hadley:
Thank you!

howardbaik:
Thank you for your work on {ellmer}!

PR #567, opened 2025-06-12
Add JSON output format to pricing script

elnelson575:
This adds a JSON output for the `prices.R` script output using `jsonlite` (already a dependency, so doesn't introduce new dependencies) and stores the JSON in `data-raw`. 

The plan would be for us to use a GitHub Action to pull this into `chatlas` whenever it's updated so we can leverage y'all's pricing work and make sure we keep the two in sync.
Let me know any thoughts or additional context needed! (Tagging in @cpsievert as well for awareness!)

### DIFF
--- a/data-raw/prices.R
+++ b/data-raw/prices.R
@@ -48,3 +48,6 @@ prices <- bind_rows(
 )
 
 usethis::use_data(prices, overwrite = TRUE, internal = TRUE)
+
+# Output JSON for ingestion into `chatlas`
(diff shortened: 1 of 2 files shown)


cpsievert:
@hadley for a bit more context -- @elnelson575 recently joined the shiny team and will be helping out a bit on `chatlas` -- welcome Liz :)

hadley:
Thanks!

PR #594, opened 2025-06-26
update req_url_path to req_url_path_append for providers

### LINKED ISSUE #593: models_ollama() strips subpaths from base_url
Our company has an Ollama server running behind a reverse proxy on a subpath URL; https://example.com/ollama. When running `ellmer::chat_ollama(base_url = "https://example.com/ollama", model = "gemma3:4b")`, I get the error:

```Error in `resp_body_json()`:
! Unexpected content type "text/html".
• Expecting type "application/json" or suffix "json".
Run `rlang::last_trace()` to see where the error occurred.```

Since version 0.2.0, ellmer::chat_ollama() [always runs ellmer::models_ollama()](https://github.com/tidyverse/ellmer/blob/87f442443471e87386799c59db90f9fbd3c7bd43/R/provider-ollama.R#L52) to check that the model is available before running. However, that function strips away the end of the base_url with req_url_path() so what should be https://example.com/ollama/api/tags results in https://example.com/api/tags

I think the issue will be resolved if references to req_url_path() are changed to req_url_path_append() in [provider-ollama.R](https://github.com/tidyverse/ellmer/blob/87f442443471e87386799c59db90f9fbd3c7bd43/R/provider-ollama.R#L117), which is consistent with the other models_*() functions in the package.

### PR DESCRIPTION
**Pull request by benyake**
Closes #593

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* `chat_aws_bedrock()`, `chat_google_gemini()`, `chat_ollama()`, and `chat_vllm()` now use a more robust method for generate model URLs from the `base_url` (#593, @benyake).
 * `Chat$chat_structured()` no longer requires a prompt (since it may be obvious from the context) (#570).
 * `models_ollama()` now includes a `capabilities` column with a comma-separated list of model capabilities (#623).
 * `chat_huggingface()` now works much better.

### CODE REVIEW

**Comment by hadley** [comment]
Would you mind use `req_url_path_relative()` here? I think it's a bit clearer.

**Comment by benyake** [comment]
`req_url_relative()` may be clearer when not using subpaths, but when subpaths are involved (which is what the issue was about) it's behavior gets very confusing, especially with slashes. This is because it's relative to the base URL which is often the domain and potentially a root path, but it's not necessarily the entire current path of the request object. Additionally, if the relative URL **starts with a slash**, it replaces the entire path after the domain. If the relative URL **does not start with a slash**, it replaces everything after the last slash in the current path. It will only append when the base URL ends in a slash and the relative URL does not.

These examples help to demonstrate. 
```
# without trailing base URL slash
req <- request("https://example.com/a/b")

req |> req_url_relative("/c/d")
<httr2_request>
GET https://example.com/c/d
Body: empty

req |> req_url_relative("c/d")
<httr2_request>
GET https://example.com/a/c/d
Body: empty

# with trailing base URL slash
req <- request("https://example.com/a/b/")

req |> req_url_relative("/c/d")
<httr2_request>
GET https://example.com/c/d
Body: empty

req |> req_url_relative("c/d")
<httr2_request>
GET https://example.com/a/b/c/d
Body: empty
```


Since base_url is the only way end users can change the location of a provider or API endpoint, using `req_url_relative()` would only resolve the subpath issue if the user explicitly includes a trailing slash in the base_url parameter, which is very prone to human error.

Even the current `models_openai()` function uses `base_url = "https://api.openai.com/v1"` as the default, so changing `req <- req_url_path_append(req, "/models")` to `req <- req_url_relative(req, "models")` would result in stripping the `/v1` to `https://api.openai.com/models`

hadley:
Hmmmm, that's a good point. It makes `req_url_relative()` less useful than I was expecting, but maybe that's appropriate — it's primarily useful when you understand both the absolute path and the relative path.

hadley:
Thanks!

PR #612, opened 2025-07-01
Use vcr for selected tests

hadley:
Which are the slowest and least reliable. This brings the total test time down to ~18s. To make further progress will need to tackle the next three slowest:

```
✔ |         29 | provider-openai [12.8s]
✔ |         37 | provider-gemini [16.3s]
✔ |        118 | chat [18.0s]
```

To do:

* [x] Consider if we need to test async tools for every provider
* [x] Update after https://github.com/ropensci/vcr/pull/527

PR #619, opened 2025-07-03
Simplify tool tests

hadley:
Although it's not super clear to me that these really need to be tested for every provider either.


hadley:
@gadenbuie just an FYI that I'm moving the tool calling tests around. This might not be their final home, but #612 suggests that we should split `test-chat.R` up into smaller pieces to benefit from parallelism.

Issue #571, opened 2025-06-13
Structured data extraction bug


hadley:
``` r
library(ellmer)

chat <- chat_anthropic()
#> Using model = "claude-3-7-sonnet-latest".
prompts <- list(
  "Kwame here - just hit the big 5-0 this year.",
  "I like apples",
  "What time is it?",
  "This cheese is 3 years old",
  "My name is Hadley."
)
type_person <- type_object(name = type_string(), age = type_number())
parallel_chat_structured(chat, prompts, type_person)
#>        name age
#> 1     Kwame  50
#> 2 <UNKNOWN>   0
#> 3 <UNKNOWN>   0
#> 4    cheese   3
#> 5    Hadley  30
```

<sup>Created on 2025-06-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hadley:
The error arises because Claude sometimes emits `{"name": "Hadley", "age": <UNKNOWN>}` (i.e. json that's not parseable). I don't see this with 4.0, so I think it's ok to close.

PR #637, opened 2025-07-09
Implement own function for converting from lists to atomics

hadley:
Biting the bullet just so I can get exactly the behaivour I want and not have to worry about weird base R edge cases. Discovered when exploring #571


R/chat-structured.R
```diff
@@ -84,3 +68,33 @@ convert_from_type <- function(x, type) {
...
+    x[types == "double"] <- lapply(x[types == "double"], as.integer)
+    types[types == "double"] <- "integer"
+  } else if (r_type == "double") {
+    x[types == "integer"] <- lapply(x[types == "integer"], as.double)
+    types[types == "integer"] <- "double"
+  }
+
+  # also captures NULLs
+  wrong_type <- types != r_type
+  x[wrong_type] <- list(`mode<-`(NA, r_type))
```
DavisVaughan: How strict do you want to be here? No warning at all for unexpected types? That feels kind of scary. I would have thought you'd handle `NULL`s explicitly up front, and then do a final check for invalid types at the end and error if any are still invalid.

R/chat-structured.R
```diff
@@ -84,3 +68,33 @@ convert_from_type <- function(x, type) {
...
+
```
DavisVaughan, line 113: What about size checks? Do you want to check that the size of each element is length 1?

R/chat-structured.R
```diff
@@ -84,3 +68,33 @@ convert_from_type <- function(x, type) {
...
+  unlist(x, use.names = FALSE)
```
DavisVaughan:
```suggestion
  unlist(x, use.names = FALSE, recursive = FALSE)
```

This may make it take a faster path too, not sure

R/chat-structured.R
```diff
@@ -84,3 +68,33 @@ convert_from_type <- function(x, type) {
...
+  }
+
+  types <- map_chr(x, typeof)
+  if (r_type == "integer") {
+    x[types == "double"] <- lapply(x[types == "double"], as.integer)
+    types[types == "double"] <- "integer"
+  } else if (r_type == "double") {
+    x[types == "integer"] <- lapply(x[types == "integer"], as.double)
+    types[types == "integer"] <- "double"
+  }
```
DavisVaughan:
```suggestion
  if (r_type == "integer") {
    doubles <- which(types == "double")
    x[doubles] <- lapply(x[doubles], as.integer)
    types[doubles] <- "integer"
  } else if (r_type == "double") {
    integers <- which(types == "integer")
    x[integers] <- lapply(x[integers], as.double)
    types[integers] <- "double"
  }
```

Depending on how much you care about performance

DavisVaughan (approved):

R/chat-structured.R
hadley: Type is already checked at the beginning?

R/chat-structured.R
hadley: I think I can assume this is correct based on the LLM following the underlying json schema.

R/chat-structured.R
DavisVaughan: I think I'm talking about what happens if your `r_type` is `"integer"` and in `x` you have a string

You want to silently convert to `NA_integer_`?

hadley: Oh yeah, I do.

hadley: Maybe I should do the same thing if they're not length 1 (i.e. silently truncate)

hadley: I did that and made it more clear that's the goal with a comment

DavisVaughan: Seems reasonable!

PR #660, opened 2025-07-15
Fix Typos in Vignettes

howardbaik:
Noticed a couple of typos in the "Getting started with ellmer" and "Tool/function calling" vignettes


hadley:
Thank you!

PR #665, opened 2025-07-16
Backup in case we need to release before vcr

hadley:


hadley:
Too hard to wind this back now; we'll just wait.

PR #667, opened 2025-07-17
Add `base_url` to `chat_aws_bedrock()`

hadley:
Includes a few other miscellanous fixes since it looks like it's been a while since I ran these tests.

Replaces #441


billsanto:
Thanks Hadley--this seems to be working for me. We will have some happy users with this change

PR #670, opened 2025-07-17
Reduce repeated code in `*_aws_bedrock()`

### LINKED ISSUE #666: HTTP 403 error running`models_aws_bedrock()` using environment variables for authentication
I have been able to successfully utilize `{ellmer}` using AWS Bedrock in my team's AWS account at the day job with `chat_aws_bedrock()`, which is a **massive win** considering most of the technical hurdles I have to overcome for these resources. To pull it off, I had to do a little customization. Our IT policies discourage the use of static API credentials for AWS services, so we have to obtain "temporary" access key ID, secret access key, and session token through another means. I have been able to obtain these "just in time" versions of the credentials thanks to linking an IAM role with a Microsoft Entra ID application. I can't share the exact details, but I can show the overall process and how it relates to the issue. First, here's the workflow I use to authenticate and set up a new chat object with `ellmer::chat_aws_bedrock()`:

```r
source("auth.R")

library(paws)
library(ellmer)

my_auth_creds <- get_creds()

# using ellmer
new_chat <- withr::with_envvar(
  new = c(
    "AWS_ACCESS_KEY_ID" = my_auth_creds$access_key_id,
    "AWS_SECRET_ACCESS_KEY" = my_auth_creds$secret_access_key,
    "AWS_SESSION_TOKEN" = my_auth_creds$session_token,
    "AWS_REGION" = "us-east-1"
  ),
  chat_aws_bedrock()
)

new_chat$chat("Tell me jokes about microsoft Windows")
```

The above approach works, since `{paws.common}` will search for environment variables related to the AWS authentication in the absence of a profile supplied. However, this same approach fails with a 403 unauthorized error if I simply try using `ellmber::models_aws_bedrock()` instead:

```r
withr::with_envvar(
  new = c(
    "AWS_ACCESS_KEY_ID" = my_auth_creds$access_key_id,
    "AWS_SECRET_ACCESS_KEY" = my_auth_creds$secret_access_key,
    "AWS_SESSION_TOKEN" = my_auth_creds$session_token,
    "AWS_REGION" = "us-east-1"
  ),
  models_aws_bedrock()
)
```

I'm trying to figure out if this is due to the IAM role used in this workflow not having the right permissions, or if it is something that `models_aws_bedrock()` is not picking up. Happy to provide more diagnostics but it won't be able to give a 100% reprex due the confidentiality of that authentication trick to get the temp session credentials.

hadley:
Maybe fixes #666

@rpodcast could you please give this ago? In principle it shouldn't change any logic, but now `models_aws_bedrock()` uses _exactly_ the same code path as `chat_aws_bedrock()`. 

But also note that the endpoints are different: `models_` uses `https://bedrock.%s.amazonaws.com` and `chat_` uses `"https://bedrock-runtime.%s.amazonaws.com"`, so that might also be your issue.

### DIFF
--- a/R/provider-aws.R
+++ b/R/provider-aws.R
@@ -60,42 +60,33 @@ chat_aws_bedrock <- function(
 ) {
   check_installed("paws.common", "AWS authentication")
   check_string(base_url, allow_null = TRUE)
-  model <- set_default(model, "anthropic.claude-3-5-sonnet-20240620-v1:0")
+  base_url <- base_url %||%
+    \(x) sprintf("https://bedrock-runtime.%s.amazonaws.com", x)
   echo <- check_echo(echo)
 
-  cache <- aws_creds_cache(profile)
-  credentials <- paws_credentials(profile, cache = cache)
-  base_url <- base_url %||%
-    paste0("https://bedrock-runtime.", credentials$region, ".amazonaws.com")
-
-  provider <- ProviderAWSBedrock(
-    name = "AWS/Bedrock",
+  provider <- provider_aws_bedrock(
     base_url = base_url,
     model = model,
     profile = profile,
-    region = credentials$region,
-    cache = cache,
     extra_args = api_args
   )
-
   Chat$new(provider = provider, system_prompt = system_prompt, echo = echo)
 }
+
 
 #' @export
 #' @rdname chat_aws_bedrock
 models_aws_bedrock <- function(profile = NULL, base_url = NULL) {
-  creds <- paws_credentials(profile)
-  base_url <- base_url %||%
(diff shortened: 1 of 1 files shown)


rpodcast:
@hadley Happy to report that it is working now!  At first, I did receive another error but it was much more transparent this time around:

```
 HTTP 403 Forbidden.
ℹ User: arn:aws:sts::123456789:assumed-role/silly-role/my-application-session is not authorized to perform: bedrock:ListFoundationModels because no identity-based policy allows the bedrock:ListFoundationModels action
```

Once I updated the associated IAM role policy to allow `bedrock:ListFoundationModels`, I was able to get the data frame of available models.

hadley:
Great, thanks!

PR #683, opened 2025-07-25
Update to latest Air settings and use `format-suggest.yaml`

DavisVaughan:
@hadley any thoughts on using `format-suggest.yaml`?
- It prevents ellmer's codebase from getting out of sync with what Air expects
- It would be good to get mileage on it before tidy dev day and ellmer seems like a good candidate

Jenny test drove it yesterday in usethis and had a pretty good experience fixing up an external pr completely within the github ui


hadley:
Sorry, I somehow missed this earlier. LGTM!

PR #704, opened 2025-08-10
Fix auth scope and API endpoints for Google Vertex

### LINKED ISSUE #523: Issue in `chat_google_vertex`: HTTP 403 Forbidden due to insufficient authentication scopes
Hi team,

I'm encountering an issue when trying to get a response from Gemini Vertex AI using the `ellmer` package. Below is the code I'm running and the error I receive:

```r
library(ellmer)

Sys.setenv(GOOGLE_APPLICATION_CREDENTIALS = "PATH_TO_JSON")

chat = chat_google_vertex(
  location = "asia-south1",
  project_id = "cifcl-analytics",
  model = "gemini-1.5-flash-002"
)

chat$chat("Hi")
```

**Error Output**
```
Error in `req_perform_connection()`:
! HTTP 403 Forbidden.
* OAuth error: insufficient_scope
* realm: https://accounts.google.com/
* scope: https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/cloud-platform.read-only https://www.googleapis.com/auth/cloud-vertex-ai.firstparty.predict
* Request had insufficient authentication scopes.
```

**Steps Taken**

* Verified that the service account has Vertex AI User permissions.
* JSON credentials are correctly referenced and readable.

**Environment**

* R version: 4.2.1
* ellmer version: 0.2.0

Thanks!

netique:
Attempts to fix #523.

- The authentication scope needed to be changed for Vertex. This PR also fixes base URL for `location = "global"`.
- Also, there is a new method (using `v1beta1` API) for getting the available models for Vertex with `models_google_vertex ()`.
- Fixes wrong subsetting of `can_generate` models in `models_google_vertex()`.

### DIFF
--- a/R/provider-google.R
+++ b/R/provider-google.R
@@ -76,7 +76,8 @@ chat_google_gemini_test <- function(
 
 #' @export
 #' @rdname chat_google_gemini
-#' @param location Location, e.g. `us-east1`, `me-central1`, `africa-south1`.
+#' @param location Location, e.g. `us-east1`, `me-central1`, `africa-south1` or
+#' `global`.
 #' @param project_id Project ID.
 chat_google_vertex <- function(
   location,
@@ -105,11 +106,28 @@ chat_google_vertex <- function(
 }
 
 # https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent
-vertex_url <- function(location, project_id) {
-  paste_c(
-    c("https://", location, "-aiplatform.googleapis.com/v1"),
+vertex_url <- function(location, project_id, list_models = FALSE) {
+  location_base <- paste0(location, "-")
+  if (location == "global") {
+    location_base <- ""
+  }
+
+  api_ver <- "v1"
+  if (list_models) {
+    api_ver <- "v1beta1"
+  }
+
+  proj_and_loc <- paste_c(
     c("/projects/", project_id),
-    c("/locations/", location),
+    c("/locations/", location)
+  )
+  if (list_models) {
+    proj_and_loc <- ""
+  }
+
+  paste_c(
+    c("https://", location_base, "aiplatform.googleapis.com/", api_ver),
+    proj_and_loc,
     "/publishers/google/"
   )
 }
@@ -587,6 +605,10 @@ default_google_credentials <- function(
 ) {
   gemini_scope <- "https://www.googleapis.com/auth/generative-language.retriever"
 
+  if (!gemini) {
+    gemini_scope <- "https://www.googleapis.com/auth/cloud-platform"
+  }
+
   check_string(api_key, allow_null = TRUE, call = error_call)
   api_key <- api_key %||% Sys.getenv("GOOGLE_API_KEY")
   if (gemini && api_key == "") {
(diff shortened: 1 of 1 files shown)


R/provider-google.R
```diff
@@ -587,6 +605,10 @@ default_google_credentials <- function(
 ) {
   gemini_scope <- "https://www.googleapis.com/auth/generative-language.retriever"
 
+  if (!gemini) {
+    gemini_scope <- "https://www.googleapis.com/auth/cloud-platform"
```
atheriel: Especially given that this is what Google's Python client is doing, I'm happy to accept that this is the "right" scope, even though it's disappointing that they don't have a finer-grained one.

R/provider-google.R
```diff
@@ -76,7 +76,8 @@ chat_google_gemini_test <- function(
 
 #' @export
 #' @rdname chat_google_gemini
-#' @param location Location, e.g. `us-east1`, `me-central1`, `africa-south1`.
+#' @param location Location, e.g. `us-east1`, `me-central1`, `africa-south1` or
```
atheriel, line 78: Does it make sense for `"global"` to be the default value for this parameter now?

R/provider-google.R
```diff
@@ -105,11 +106,28 @@ chat_google_vertex <- function(
 }
 
 # https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent
-vertex_url <- function(location, project_id) {
-  paste_c(
-    c("https://", location, "-aiplatform.googleapis.com/v1"),
+vertex_url <- function(location, project_id, list_models = FALSE) {
```
atheriel: I think this has become too convoluted in trying to support the model listing endpoint. You could move that logic into the `models_google_vertex()` function below instead.

R/provider-google.R
```diff
@@ -687,20 +709,29 @@ models_google_gemini <- function(
 
   json <- resp_body_json(resp)
 
+  if (isFALSE(api_key)) { # if Vertex
+    name <- map_chr(json$publisherModels, "[[", "name")
+    name <- gsub("^publishers/google/models/", "", name)
+    # this is the closest to "generateContent" in "supportedGenerationMethods" for Gemini
+    # https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/publishers.models
+    can_generate <- json$publisherModels |> map_lgl(\(x) "openGenerationAiStudio" %in% names(x$supportedActions))
+  } else {
```
atheriel, line 743: Nit: I think the indentation is off here now. Can you run this through `air` (if you have it).

R/provider-google.R
netique: I don't think we can set any default here – for the Python client, you have to provide both project ID _and_ location, unless you use the API key. They don't specify any default value.

R/provider-google.R
netique: I completely agree, this is too much for my taste ;-)

R/provider-google.R
netique: Sure, will do.  I was using tidyverse styler.

R/provider-google.R
hadley: It'd be useful if you could add a comment pointing to the python source here too. (Just so we have a link back to check if anything changes)

netique:
Many thanks for the review! I think the comments should all be resolved now, news bullets were added as well. Maybe one thing to check is the `project_id` deprecation in `models_google_vertex()`. I'm not sure what the policy is, but I used `{lifecycle}` for that (you have to update the version for the release).

hadley:
@netique I did a bit more refactoring to make the gemini and vertex paths more clear. Could you please double check I haven't broken anything on the vertex side?

netique:
@hadley Thanks! This is much clearer now. I always hesitate a bit to make any bigger changes in PRs, maybe I shouldn't ;-)

The only thing to fix was a small typo, but it seems that the model listing API on the Google side has changed somehow since my last commit... Now it requires an additional header with a project ID, which was completely superfluous before, now it's mandatory – but not in the URL. I suspect getting the list of available models is going to be tricky since they keep changing it, and it's not even supported in the Python client.

hadley:
Thanks for the contributions! For important stuff like this, we're always happy to provide feedback on the implementation, so don't feel bad about bigger changes.

PR #724, opened 2025-08-21
convert png logo to svg

chendaniely:
used the https://www.freeconvert.com/png-to-svg/download tool (it was better than the adobe version)


hadley:
Can you remind me why we want to do this?

chendaniely:
the svg was mostly to help with larger print outs of the logo. For example some of the larger banners (e.g., during scipy) was slightly blurry. we should re-confirm if we make any more large banners what the quality would be on the final print out before fully accepting any SVG converted image.

hadley:
Ok, in that case https://github.com/rstudio/hex-stickers would be the place to start.

PR #729, opened 2025-08-24
Fix NEWS bullet for already merged PR #704

netique:
Due to last-minute changes on Google's side, the `project_id` argument is again mandatory, as it was before, but this hasn't been reflected in the NEWS.


hadley:
Something got messed up here so I'll do by hand.

PR #733, opened 2025-08-26
Fix #732 by de-duplicating duplicated JSON responses from AWS Bedrock in structured data extraction

### LINKED ISSUE #732: AWS Bedrock provider intermittently creates duplicate ContentJson objects, causing extraction failures
## Problem Summary

The AWS Bedrock provider sometimes creates duplicate ContentJson objects from a single model response, causing `extract_data()` to fail with "Data extraction failed: 2 data results received."

## Environment
   
- **Provider:** AWS Bedrock (`chat_aws_bedrock`)
- **Model:** `us.meta.llama4-maverick-17b-instruct-v1:0` (likely affects other Bedrock models)
- **Function:** `parallel_chat_structured()` and `chat_structured()`

## Expected Behavior

`extract_data()` should receive exactly 1 ContentJson object per structured response.

## Actual Behavior

Intermittently (non-deterministically), `extract_data()` receives 2 identical ContentJson objects from the same response, causing extraction to fail.

## Reproduction

The issue is intermittent but can be reproduced by running structured extraction multiple times. I have worked out a fix, will post tests that recreate the error there. But generically,

  ```r
  # Setup
  chat <- chat_aws_bedrock(
    model = "us.meta.llama4-maverick-17b-instruct-v1:0",
    system_prompt = "Extract structured data...",
    api_args = list(inferenceConfig = list(seed = 26082025, temperature = 0))
  )

  # Run multiple times - failures occur randomly
  for (i in 1:10) {
    tryCatch({
      result <- chat$chat_structured(some_text, type = some_type)
    }, error = function(e) {
      if (grepl("2 data results", e$message)) {
        cat("Failure on attempt", i, "\n")
      }
    })
  }
```

##  Investigation Results

-  All failure cases show exactly 2 identical ContentJson objects (not different responses)
-  Success cases show exactly 1 ContentJson object
-  The issue appears to be in the Bedrock provider's response parsing, not the model output
-  Non-deterministic despite fixed seed, suggesting a parsing race condition or API state issue

##  Impact

-  Causes `parallel_chat_structured()` batch jobs to fail partway through

##  Proposed Solution

Modify `extract_data()` to handle duplicate ContentJson objects by:
1. Detecting when n=2 JSON objects are identical
2. Using the first object and issuing a warning
3. Still failing if JSON objects are genuinely different (unexpected case)

### PR DESCRIPTION
**Pull request by kbenoit**
## Problem

AWS Bedrock occasionally returns duplicate identical JSON objects during structured data extraction, causing
`extract_data()` to fail with the error:
```
  Data extraction failed: 2 data results received.
```
This causes `parallel_chat_structured()` to fail completely when processing batches of prompts.

## Evidence

See also the reprex based on my own data, in https://github.com/tidyverse/ellmer/issues/732#issuecomment-3222467503.

Captured failure case from real-world usage (Spanish political manifesto analysis):
  
```r
# The Turn object contains two identical ContentJson objects:
turn@contents:
  [[1]] <ellmer::ContentJson>
  @value: List of 9
$ Country_LLM         : chr "Spain"
$ Party_LLM           : chr "Partido Popular"
$ is_populist         : logi FALSE
$ is_populist_evidence: chr "The party does not consistently frame politics in populist terms..."
$ antielitism         :List of 2
..$ evidence: chr "The party criticizes the current government's policy of 'cesiones humillantes'..."
..$ score   : int 4
$ generalwill         :List of 2
..$ evidence: chr "The party emphasizes the importance of unity and the general will of the people..."
..$ score   : int 5
$ indivisible         :List of 2
..$ evidence: chr "The party portrays Spain as a single nation with a common identity..."
..$ score   : int 6
$ manichean           :List of 2
..$ evidence: chr "The party criticizes the current government's policies and actions..."
..$ score   : int 3
$ peoplecentrism      :List of 2
..$ evidence: chr "The party emphasizes the importance of listening to the people..."
..$ score   : int 5

[[2]] <ellmer::ContentJson>
  @value: List of 9

# IDENTICAL structure and values as [[1]]
# Verification: identical(json1, json2) == TRUE
```
Scale: Captured 11 failure cases vs 3 success cases during batch processing, indicating this is a significant
intermittent issue with Bedrock's tool calling mechanism.

  Solution

  Updated extract_data() in R/chat-structured.R to gracefully handle multiple JSON responses:

  if (n == 2) {
    # Check if the two JSON objects are identical (duplicate case)
    if (identical(val1, val2)) {
      warning("Found duplicate JSON responses, using the first one", call. = FALSE)
      out <- val1
    } else {
      # Different JSON objects - use the last one (likely the final response)  
      warning("Found multiple different JSON responses, using the last one", call. = FALSE)
      out <- val2
    }
  }

## Tests

  Added comprehensive tests in test-chat-structured.R:
  - Handle duplicate identical JSON responses with warning
  - Handle different JSON responses (use last one)
  - Include prompt index in warning messages for parallel operations
  - Error appropriately on >2 JSON responses

##  Impact

Before:
```r
# This would fail completely
parallel_chat_structured(chat, prompts, type = my_type)
#> Error: Data extraction failed: 2 data results received.
```

After:
```r
# This now succeeds with warning
result <- parallel_chat_structured(chat, prompts, type = my_type)
#> Warning: Found duplicate JSON responses, using the first one (prompt 43).
#> ... successful extraction continues

This ensures robust structured data extraction for production use, even when providers exhibit intermittent duplication behaviour.

## Additional notes

This works best when combined with #725. #705 does not fix this issue, but if the other two PRs are accepted, I will redo #705 and integrate them into a single, more `robust parallel_chat_structured()`.

### DIFF
--- a/R/chat-structured.R
+++ b/R/chat-structured.R
@@ -1,12 +1,37 @@
-extract_data <- function(turn, type, convert = TRUE, needs_wrapper = FALSE) {
+extract_data <- function(turn, type, convert = TRUE, needs_wrapper = FALSE, prompt_index = NULL) {
   is_json <- map_lgl(turn@contents, S7_inherits, ContentJson)
   n <- sum(is_json)
-  if (n != 1) {
-    cli::cli_abort("Data extraction failed: {n} data results recieved.")
+  
+  if (n == 0) {
+    cli::cli_abort("Data extraction failed: 0 data results received.")
+  } else if (n == 1) {
+    # Normal case - exactly 1 JSON object
+    json <- turn@contents[[which(is_json)]]
+    out <- json@value
+  } else if (n == 2) {
+    # Check if the two JSON objects are identical (duplicate case)
+    json_indices <- which(is_json)
+    json1 <- turn@contents[[json_indices[1]]]
+    json2 <- turn@contents[[json_indices[2]]]
+    
+    val1 <- json1@value
+    val2 <- json2@value
+    
+    if (identical(val1, val2)) {
+      # Duplicate case - use the first one
+      index_msg <- if (!is.null(prompt_index)) paste0(" (prompt ", prompt_index, ")") else ""
+      warning("Found duplicate JSON responses, using the first one", index_msg, ".", call. = FALSE, immediate. = TRUE)
+      out <- val1
+    } else {
+      # Different JSON objects - use the last one (likely the final response)
+      index_msg <- if (!is.null(prompt_index)) paste0(" (prompt ", prompt_index, ")") else ""
+      warning("Found multiple different JSON responses, using the last one", index_msg, ".", call. = FALSE, immediate. = TRUE)
+      out <- val2
+    }
+  } else {
+    # More than 2 JSON objects - this is unexpected
+    cli::cli_abort("Data extraction failed: {n} data results received. Expected 1 or 2.")
   }
-
-  json <- turn@contents[[which(is_json)]]
-  out <- json@value
 
   if (needs_wrapper) {
     out <- out$wrapper
--- a/R/parallel-chat.R
+++ b/R/parallel-chat.R
@@ -170,12 +170,13 @@ multi_convert <- function(
 ) {
   needs_wrapper <- type_needs_wrapper(type, provider)
 
-  rows <- map(turns, \(turn) {
+  rows <- imap(turns, \(turn, idx) {
     extract_data(
       turn = turn,
       type = wrap_type_if_needed(type, needs_wrapper),
       convert = FALSE,
-      needs_wrapper = needs_wrapper
+      needs_wrapper = needs_wrapper,
+      prompt_index = idx
     )
   })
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by hadley** [comment]
I think a more general approach here would be to just warn if there are more than two `ContentJson`s and then just use the first one. I think the original code existed out of an abundance of caution, and I don't have any evidence that it's actually important. Do you want to have a go at implementing that? I'm also happy to do it if you don't have the time.

**Comment by kbenoit** [comment]
I'm happy to have a go at this, once the other parallel error handling PRs have been merged into main, and I can update this branch.

**Comment by kbenoit** [comment]
> I think a more general approach here would be to just warn if there are more than two `ContentJson`s and then just use the first one. I think the original code existed out of an abundance of caution, and I don't have any evidence that it's actually important. Do you want to have a go at implementing that? I'm also happy to do it if you don't have the time.

Implemented in bc75491 (although there was something to be said for the more informative version for handling duplicated pairs or n > 2)

**Comment by hadley** [comment]
I've simplified your implementation back to the bare minimum — I think it's better to start there and get more complicated as we need it as I don't think we should let buggy implementations overly guide the design of ellmer.  If you do think informing the user about exactly which prompt had the problem, I think it would be better to attack the problem by wrapping the conditions, like this:

```R
withCallingHandlers(
  warn("hi"),
  warning = function(w) {
    warn("bye", parent = w)
    invokeRestart("muffleWarning")
  }
)
```

That would also be helpful for errors too (assuming that it's actually the prompt that's causing the problem).

PR #756, opened 2025-09-20
Update naming claude brand in documentation, `document()`

xx02al:
According to Anthropic's announcement earlier this week: Today, we’re unifying all of our developer offerings under the Claude brand. You'll see updated naming across our platform and documentation, but your technical implementation stays the same—no action needed.


hadley:
Thanks!

Issue #743, opened 2025-09-08
`batch_chat_*()` updates `token_usage()` on every call


hadley:
Even if it's just updating previous results.

Might also be worth providing `batch_chat_price()` ?

hadley:
* Extract out `value_tokens()` generic with methods for each provider
* Add `log_tokens = TRUE` argument to `value_turn()`, and ensure all methods respect it
* Implement `batch_results_token()` generic + methods
* Implement `batch_chat_price()`

PR #784, opened 2025-10-15
Improvements to token API

hadley:
Most of the work was done by Claude code, but it was all closely supervised by me. Initial refactoring to support #743.

@gadenbuie I'm not sure if I should bite the bullet and create a Token class instead of just using a named vector.

### DIFF
--- a/R/tokens.R
+++ b/R/tokens.R
@@ -15,10 +15,6 @@ tokens <- function(input = 0, output = 0, cached_input = 0) {
 }
 
 tokens_log <- function(provider, tokens) {
-  if (is.null(tokens)) {
-    return(invisible())
-  }
-
   i <- tokens_match(
     provider@name,
     provider@model,


gadenbuie (approved):
These changes all look good to me!

I'm equally on the fence about creating a `Tokens` class; the named vector gives me pause just because subsetting throws if a field is missing. Would it make sense to use a list here so that you can use `tokens[["cached_input"]] %||% 0` or `tokens[["cached_input"]] %||% NA`?

A `Token` class would help with validation and default values, and it'd be nice to have if those need to be enforced. Otherwise a simple data structure seems fine here.

For my own curiosity and education: how/where is the token log used (`the$tokens`)?

hadley:
Maybe I should just compromised for now and make it a named list that's always constructed by a single function. 

`the$tokens()` is used in `token_usage()` which gives you the total usage for the current session.

R/provider-anthropic.R
```diff
@@ -296,14 +304,9 @@ method(value_turn, ProviderAnthropic) <- function(
...
-  tokens <- tokens_log(
-    provider,
-    input = result$usage$input_tokens,
-    output = result$usage$output_tokens,
-    cached_input = result$usage$cache_read_input_tokens
-  )
-
-  assistant_turn(contents, json = result, tokens = tokens)
+  tokens <- value_tokens(provider, result)
+  tokens_log(provider, tokens)
```
cpsievert, line 308: @hadley seems this part could be pulled out into `chat.R`? That way the provider implementation won't be responsible for the logging?

hadley: Where exactly are you thinking? I agree that it would good to make it more systematic.

cpsievert: Just before storing the turns? https://github.com/tidyverse/ellmer/blob/ab18d5ff/R/chat.R#L703

hadley: Yeah, that sounds reasonable. Can you please file an issue so I don't forget?

PR #785, opened 2025-10-15
Describe the schema of a data frame

### LINKED ISSUE #744: Function to describe a data frame
This is needed in lots of places and is a bit fiddly to get right. I think it's probably worth having something in ellmer.

hadley:
I've ensured that various special cases (e.g. zero or one rows) don't error, but haven't otherwise tried to optimise their display.

Fixes #744

@simonpcouch @gadenbuie @jcheng5 let me know if there's anything else that you think would be useful here. I'd also love thoughts on the name. Presuambly in the future we might also have a `df_data()`/`df_raw()` or similar that would ship an entire data frame to the LLM in the format we believe to be most useful.

### DIFF
--- a/R/schema.R
+++ b/R/schema.R
@@ -12,11 +12,13 @@
 #' conciseness and accuracy.
 #'
 #' @param df A data frame to describe.
+#' @param max_cols Maximum number of columns to includes. Defaults to 50 to
+#'   avoid accidentally generating very large prompts.
 #' @export
 #' @examples
 #' df_schema(mtcars)
 #' df_schema(iris)
-df_schema <- function(df) {
+df_schema <- function(df, max_cols = 50) {
   if (!is.data.frame(df)) {
     stop_input_type(df, "a data frame")
   }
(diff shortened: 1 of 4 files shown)


R/schema.R
```diff
@@ -0,0 +1,179 @@
...
+    stop_input_type(df, "a data frame")
+  }
+
+  df_desc <- sprintf(
+    "A data frame with %i rows and %i columns:",
+    nrow(df),
+    ncol(df)
+  )
+
+  cols <- map_chr(df, col_schema)
```
simonpcouch, line 40: I do wonder if we ought to subset to the first $n$ columns here in cases of very wide data. 20? 50?

hadley:
@simonpcouch I should have mentioned that I wrote this with an eye to being standalone so that (e.g.) positron assistant could also easily use it.

R/schema.R
hadley: Maybe have a threshold and warn when above it? I think I'd lean more towards 50 than 20.

Issue #628, opened 2025-07-08
Add error handling to `parallel_chat`


paddytobias:
Thanks for this package, it's been great! 

I'm using the Azure OpenAI service with `parallel_chat` as the batch processing isn't currently available. However for some inexplicable reason, the batch of prompts I send contain some records that trigger a 4xxx response due to the content filtering policy of [Azure OpenAI](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/content-filter). 

I'd like to suggest adding error handling in `parallel_chat` so the whole process doesn't fail if some errors are returned. `parallel_chat` is calling `httr2::req_perform_parallel` internally which has an `on_error` param for error handling. The easiest solution is probably just adding `...` to the function that get passed to `httr2::req_perform_parallel`, then all params will be available in `parallel_chat`. 

I've made the update in `parallel_chat` and `parallel_chat_structured` and happy to open up a PR, but currently don't have permission.

hadley:
What would you put in the output for inputs that errored? For `parallel_chat()`, I think it would probably be ok to mix Chat and error objects, but what about `parallel_chat_structured()`?

kbenoit:
First, let me say, `parallel_chat_structured()` is an absolutely game-changing function for the sort of work I've been applying this to. so: 🙌!!   I recently started working with it, after writing tedious code to try to vectorise `chat_structured()` based on earlier versions of the package, but also with a lot of error handling, since I very commonly encounter errors due to strange texts, JSON parsing glitches, hitting rate limits, data types that do not conform to the `type` specs, and -- perhaps most common -- exceeding the token window sizes. 

`parallel_chat_structured()` is better than `chat_structured()` since it subsumes it, when you specify `max_active = 1`, useful when running ollama models on limited machines. It's also better since it structures the output as a data.frame by default and since it includes the tokens in and out as well as the cost 👏👏👏.

Back to the error question, I've created a reproducible example below including a link to the data I fed it to create the token limit error.

How would I put this into the output? How about:

`include_status = FALSE` in the function signature

If `TRUE`, then record `""` for elements that succeed, and the error message for inputs that fail, for a `.status` column.

For the field values that match the flattened `type` definitions, record `“”`, NULL, or NA as appropriate to the object type, when `.status != “”`. 

Reprex:

``` r
library("ellmer")

# get some UK 2010 party manifestos I've been working with

[...]

PR #787, opened 2025-10-16
Improve error handling in `parallel_chat_structured()`

hadley:
Now it will warn, instead of failing, if some results are not parseable. This change includes making `ContentJson()` lazier at parsing JSON so we can during data extraction, not object creation. Also includes some tests (and a small fix) to ensure that `NULL` can be coerced to `NA`s of the correct type.

Part of #628

@kbenoit I think this resolves the problems related to data extraction, and makes sense to me as a self-contained PR. Note that from my exploration is looks like `convert_from_type()` already did most of the work needed to generate missing responses provided that you feed it `NULL`.

### DIFF
--- a/R/content.R
+++ b/R/content.R
@@ -314,14 +314,14 @@ ContentJson <- new_class(
   "ContentJson",
   parent = Content,
   properties = list(
-    # Some providers guarantee JSON returning it as part of the parsed results
-    # others return as a string, which may or may not be valid JSON. We don't
-    # want to force conversion in value_turn() since that might error, which
-    # happens rather too soon.
+    # Some providers guarantee JSON returning it as part of the parsed results.
+    # Others return as a string, which may or may not be valid JSON. We don't
+    # want to force conversion on construction() since that's likely to be
+    # inconveniently early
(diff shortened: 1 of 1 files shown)


kbenoit:
Thanks @hadley. But it still fails for the example in https://github.com/tidyverse/ellmer/issues/628#issuecomment-3096621553 where one of the input texts exceeds the model's input token length. Reprex in in #628 but the failing output, reproduced here, is

```r
> result_fail <- parallel_chat_structured(
+     prompts = as.list(data_manifestos),
+     chat = chat,
+     type = immig_type
+ )
Error in `req_perform_parallel()`:
! HTTP 400 Bad Request.
ℹ This model's maximum context length is 128000 tokens. However, your messages resulted in 174910 tokens (including 143 in the response_format
  schemas.). Please reduce the length of the messages or schemas.
Run `rlang::last_trace()` to see where the error occurred.

> result_fail
Error: object 'result_fail' not found
```

hadley:
@kbenoit yup, that's a separate problem which I'm working on now in a separate PR.

PR #789, opened 2025-10-16
Improve error handling for `parallel_chat_*()`

### LINKED ISSUE #628: Add error handling to `parallel_chat`

hadley:
Add a `on_error` argument to match `httr2::req_parallel_perform()` and default it to "continue". Then ensure that all output types revealed errored and cancelled requests appropriately.

Fixes #628

@kbenoit this is the final piece

### DIFF
--- a/man/parallel_chat.Rd
+++ b/man/parallel_chat.Rd
@@ -87,8 +87,8 @@ prompt. Requests that weren't succesful get an \code{NA}.
 For \code{parallel_chat_structured()}, a single structured data object with one
 element for each prompt. Typically, when \code{type} is an object, this will
 be a tibble with one row for each prompt, and one column for each
-property. If the output is a data frame, and some requests errors,
-an \code{.error} column will be added with the error messages.
+property. If the output is a data frame, and some requests error,
+an \code{.error} column will be added with the error objects.
 }
 \description{
 \ifelse{html}{\href{https://lifecycle.r-lib.org/articles/stages.html#experimental}{\figure{lifecycle-experimental.svg}{options: alt='[Experimental]'}}}{\strong{[Experimental]}}


kbenoit:
Some issues with the return object could be tricky. The info is there, but the object breaks the standard data.frame methods (like print).

``` r
library("ellmer")

# get some texts
url <- "https://www.dropbox.com/scl/fi/bwp57e16zt3dbiq7c0vd3/data_manifestos.rda?rlkey=hpo889hfhzpcjro3sbz3e743g&dl=1"
temp <- tempfile(fileext = ".rda")
download.file(url, destfile = temp, mode = "wb")  # mode = binary write
load(temp)

# make the Labour file exceed the 128k context window
data_manifestos["Lab_2010"] <- paste(
  rep(data_manifestos["Lab_2010"], 4),
  collapse = "\n\n---\n\n"
)

immig_type <- type_object(
  party = type_string("The name of the political party."),
  immig_score = type_integer("An integer position from -1 to 1 on immigration and
                             asylum seekers, where:
                             -2 means strongly against any immigration,
                             -1 means moderately opposed,
                             0 means neutral,
                             1 means moderately in favor of a more permissive stance on immigration,
                             2 means strongly in favour of a permissive stance on immmigration."),
  immig_rationale = type_string("A brief rationale for your answer.")
)

chat <- chat_openai(
  model = "gpt-4o",
  system_prompt = "Extract structured data from political texts.  You will be
  scoring the position of each party on immigration and asylum seekers."
)

result_fail <- result_fail <- parallel_chat_structured(
  prompts = as.list(data_manifestos),
  chat = chat,
  type = immig_type,
  on_error = "continue"
)
#> [working] (0 + 0) -> 8 -> 1 | ■■■■                              11%
#> [working] (0 + 0) -> 6 -> 3 | ■■■■■■■■■■■                       33%
#> [working] (0 + 0) -> 0 -> 9 | ■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■  100%
#> Warning: 1 request errored.

library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
glimpse(result_fail)
#> Rows: 9
#> Columns: 4
#> $ party           <chr> "The Coalition", "British National Party", "Conservati…
#> $ immig_score     <int> -1, -2, -1, 1, NA, 1, 1, 1, -2
#> $ immig_rationale <chr> "The Coalition's stance on immigration emphasizes cont…
#> $ .error          <list> <NULL>, <NULL>, <NULL>, <NULL>, [<error/httr2_http_400…
glimpse(select(result_fail, party:immig_rationale))
#> Rows: 9
#> Columns: 3
#> $ party           <chr> "The Coalition", "British National Party", "Conservati…
#> $ immig_score     <int> -1, -2, -1, 1, NA, 1, 1, 1, -2
#> $ immig_rationale <chr> "The Coalition's stance on immigration emphasizes cont…

class(result_fail)
#> [1] "data.frame"
print.data.frame(result_fail)
#> Error in format.default(unlist(xx), ...): Found no format() method for class "weakref"
```

<sup>Created on 2025-10-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/parallel-chat.R
```diff
@@ -54,7 +66,13 @@
 #' type_person <- type_object(name = type_string(), age = type_number())
 #' parallel_chat_structured(chat, prompts, type_person)
 #' \dontshow{ellmer:::vcr_example_end()}
-parallel_chat <- function(chat, prompts, max_active = 10, rpm = 500) {
+parallel_chat <- function(
+  chat,
+  prompts,
+  max_active = 10,
+  rpm = 500,
+  on_error = "return"
```
kbenoit: Would this be better matching the structure of `httr2::req_perform_parallel()`, as in:

```
#' @param on_error Character string specifying how to handle errors:
#'   `"continue"` (default) continues processing all prompts, returning NA for failed ones (maintains input length),
#'   `"return"` stops at first error and returns only successful results processed before the error, printing the error message,
#'   `"stop"` stops on first error like the original function.
```

(from https://github.com/tidyverse/ellmer/pull/705/files#diff-3c0ac75881c575e3ff5a68fc8b9b54c8c012f4dbd5b6986d02f9f12daae2dc69R193)

hadley:
@kbenoit yes, I'm becoming increasingly convinced that we just need to bite the bullet and return tibbles here.

R/parallel-chat.R
hadley: Yes — that's what I get for having claude implement this bit 😬

hadley:
@simonpcouch could you please take a look? Can you please contemplate if there are any downsides to returning tibbles here?

R/parallel-chat.R
```diff
@@ -74,40 +93,71 @@ parallel_chat <- function(chat, prompts, max_active = 10, rpm = 500) {
...
+    prompts,
+    max_active = max_active,
+    rpm = rpm,
+    on_error = on_error
+  )
+
+  is_ok <- !map_lgl(chats, turn_failed)
+  out <- rep(NA_character_, length(prompts))
+  out[is_ok] <- map_chr(chats[is_ok], \(chat) chat$last_turn()@text)
+  out
```
simonpcouch, line 160: This was my main question when I see the `"return"` default—glad the result is always the same length!

R/parallel-chat.R
```diff
@@ -25,13 +25,24 @@
...
+#'
+#' For `parallel_chat_text()`, a character vector with one element for each
+#' prompt. Requests that weren't succesful get an `NA`.
+#'
 #' For `parallel_chat_structured()`, a single structured data object with one
 #' element for each prompt. Typically, when `type` is an object, this will
-#' will be a data frame with one row for each prompt, and one column for each
-#' property.
+#' be a tibble with one row for each prompt, and one column for each
+#' property. If the output is a data frame, and some requests errors,
```
simonpcouch:
```suggestion
#' property. If the output is a data frame, and some requests error,
```

R/parallel-chat.R
```diff
@@ -25,13 +25,24 @@
...
+#' an `.error` column will be added with the error messages.
```
simonpcouch:
```suggestion
#' an `.error` column will be added with the error objects.
```

When I first read this, my vitals / tune spidey sense was that we would probably want to go ahead and return the error objects themselves rather than just the messages. After running locally, though, looks like this is actually the case!

simonpcouch (approved):
No concerns about switching to tibbles. LGTM!

Issue #712, opened 2025-08-12
Support openai flex pricing


bshor:
Hello -- upon the heels of ellmer's support for batch APIs, I would love (without any expectations of timing!) for support for  "flex" pricing, which seems closer to normal API requests, but at batch pricing levels.

<img width="743" height="493" alt="Image" src="https://github.com/user-attachments/assets/b95e83ee-82ef-4b92-b930-1383d79b00a4" />

AFAIK this doesn't yet exist in ellmer.

hadley:
Can you please provide a link to this page?

bshor:
Of course. Here is the [documentation of flex processing](https://platform.openai.com/docs/guides/flex-processing?api-mode=responses) from OpenAI and here's the [pricing page](https://platform.openai.com/docs/pricing?latest-pricing=flex).

hadley:
Two parts to this. The easy part is adding an new parameter to `chat_openai()` to select the service level. The harder (but necessary!) part is to add some sort of additional "discount" to token pricing, which we also need for batch tokens. Probably should do after #743.

PR #790, opened 2025-10-16
Add pricing variants

hadley:
Not used anywhere yet, but the data and logic is in place. The main downside of this approach is that LiteLLM data doesn't appear to have "batches" data for many models, so I'd need to add this.

Part of #712

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* Updated pricing data (#790).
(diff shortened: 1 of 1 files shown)


hadley:
Hmmm, actually it's probably easier for me to just add the batches pricing since it's always 50% off currently — and we only need to do it for the supported providers which is only claude and openAI currently.

cpsievert (approved):

PR #793, opened 2025-10-19
Fix on_error='return' behavior in parallel_chat functions

kbenoit:
This is a PR to the branch that is currently in #789 - a PR into a PR.

@hadley I tested the `on_error = "return"` behaviour and found it wasn't working as expected - it was returning all results just like `on_error = "continue"`. I used my reprex code from https://github.com/tidyverse/ellmer/issues/628#issuecomment-3096621553 with testing the various `on_error` options. In that test, the fifth input fails because it exceeds the max context window.

The issue is that when requests are executed in parallel with `max_active = 10`, even if request number 5 fails, requests 6-9 still complete successfully. The fix needed was to filter results in `multi_convert()` to keep only results up to and including the first error.

I've pushed a fix to my fork that properly implements the "stop at first error" behavior:
- Commit: https://github.com/kbenoit/ellmer/commit/359a142
- Branch: https://github.com/kbenoit/ellmer/tree/parallel-on-error

The fix adds logic in `multi_convert()` to truncate results at the first error when `on_error = "return"`, and includes tests to verify the behaviour. It seems a bit drastic (and wasteful) to truncate the results but it does provide a behaviour that is now as advertised, which is "return: stop processing and return all responses so far, including any errors."

Changes made:

1. **Modified `multi_convert()` in `R/parallel-chat.R`**:
   - Find the first error in the results
   - Keep only results up to and including that error
   - Discard any results after the first error, even if they succeeded

2. **Added early return logic in `parallel_chat()`**:
   - Stop processing tool calls when `on_error = "return"` and an error occurs

3. **Added comprehensive tests in `test-parallel-chat.R`**:
   - Verify `on_error = "return"` stops at first error
   - Verify `on_error = "continue"` processes all requests
   - All existing tests still pass (48 tests in parallel-chat suite)

I also merged the latest changes from main - these were hung up just after [fe9dff9](https://github.com/tidyverse/ellmer/pull/789/commits/fe9dff99810858fabe3182d98c21d1b3ef39517a) because of merge conflicts, but those were just regarding NEWS.md updates.

### DIFF
--- a/tests/testthat/test-parallel-chat.R
+++ b/tests/testthat/test-parallel-chat.R
@@ -225,15 +225,36 @@ test_that("on_error = 'return' in parallel_chat_structured stops at first error"
   # Mock responses for on_error = "return": 2 successes, 1 error, then 2 more successes
   # The last 2 successes should be filtered out when on_error = "return"
   responses_return <- list(
-    Turn("assistant", list(ContentJson(list(name = "Alice", age = 30))), tokens = c(10, 20, 0)),
-    Turn("assistant", list(ContentJson(list(name = "Bob", age = 25))), tokens = c(10, 20, 0)),
+    Turn(
+      "assistant",
+      list(ContentJson(list(name = "Alice", age = 30))),
+      tokens = c(10, 20, 0)
+    ),
+    Turn(
+      "assistant",
+      list(ContentJson(list(name = "Bob", age = 25))),
+      tokens = c(10, 20, 0)
+    ),
     simpleError("Request failed"),
-    Turn("assistant", list(ContentJson(list(name = "Charlie", age = 35))), tokens = c(10, 20, 0)),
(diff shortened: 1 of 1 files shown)


hadley:
If a bunch of requests have started and one fails, I think you might as well wait for all the others to return (since you've paid for them). That's the point of `on_error = "return"`. The difference with `on_return = "continue"` is that it will continue to add new requests to the queue. Does that make sense?

kbenoit:
Yes, makes sense, and does fit with a general interpretation of "return all responses so far". 

Then reverting [359a142](https://github.com/tidyverse/ellmer/pull/793/commits/359a1426f192e45689567f56f36a5b45ce8c5634) also makes sense.

kbenoit:
Closing this PR since then the only diff with [tidyverse:parallel-on-error](https://github.com/tidyverse/ellmer/tree/parallel-on-error) in #789 is a trivial conflict resolution in NEWS.md.

PR #798, opened 2025-10-21
add `@duration` slot to turns

### LINKED ISSUE #479: `@completed` values are shared among turns submitted in parallel
In the below reprex, the 20 prompts are submitted at different times and a response finishes streaming at different times, but all of the user turns have the same `@completed` value, and the same goes for the assistant turns.

``` r
library(ellmer)

ch <- chat_openai(model = "gpt-4.1-nano")

# reduce max_active to ensure turns are submitted gradually
res <- ch$chat_parallel(as.list(rep("hi", 20)), max_active = 2)
#> [working] (12 + 0) -> 2 -> 6 | ■■■■■■■■■■                        30%
#> [working] (2 + 0) -> 2 -> 16 | ■■■■■■■■■■■■■■■■■■■■■■■■■         80%
#> [working] (0 + 0) -> 0 -> 20 | ■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■  100%

vapply(res, function(chat) chat$get_turns()[[1]]@completed, Sys.time())
#>  [1] 1746648220 1746648220 1746648220 1746648220 1746648220 1746648220
#>  [7] 1746648220 1746648220 1746648220 1746648220 1746648220 1746648220
#> [13] 1746648220 1746648220 1746648220 1746648220 1746648220 1746648220
#> [19] 1746648220 1746648220
vapply(res, function(chat) chat$last_turn()@completed, Sys.time())
#>  [1] 1746648223 1746648223 1746648223 1746648223 1746648223 1746648223
#>  [7] 1746648223 1746648223 1746648223 1746648223 1746648223 1746648223
#> [13] 1746648223 1746648223 1746648223 1746648223 1746648223 1746648223
#> [19] 1746648223 1746648223
```

<sup>Created on 2025-05-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

With this progress bar in mind: `[working] (a + b) -> c -> d`.

...I would have expected that each user turn's value `chat$get_turns()[[1]]@completed` reflected when that turn _last_ transitioned into the `c` slot (as in, if it hit a rate limit, the user turn would reflect the last time when the turn was submitted). I would have expected that each assistant turn's value `chat$last_turn()@completed` reflected when that chat was transitioned into the `d` slot.

simonpcouch:
Closes #479.

With this PR:

``` r
library(ellmer)

ch <- chat_openai(model = "gpt-4o-mini")

# reduce max_active to ensure turns are submitted gradually
res <- parallel_chat(ch, as.list(rep("hi", 20)), max_active = 2)
#> [working] (14 + 0) -> 2 -> 4 | ■■■■■■■                           20%
#> [working] (7 + 0) -> 2 -> 11 | ■■■■■■■■■■■■■■■■■                 55%
#> [working] (0 + 0) -> 0 -> 20 | ■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■  100%

# user turns are NA
vapply(res, function(chat) chat$get_turns()[[1]]@duration, numeric(1))
#>  [1] NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA

# each assistant turn has a timing extracted from httr2
vapply(res, function(chat) chat$last_turn()@duration, numeric(1))
#>  [1] 0.856290 0.619388 0.776574 0.609884 0.704185 0.685017 0.420721 0.384669
#>  [9] 0.410552 0.566716 0.702766 0.554898 0.551356 0.656204 0.622301 0.588885
#> [17] 0.385447 0.840800 0.490942 0.478413

# same goes for sequential chats
ch$chat("hi", echo = FALSE)
#> Hello! How can I assist you today?
vapply(ch$get_turns(), function(turn) turn@duration, numeric(1))
#> [1]       NA 0.457641
```

<sup>Created on 2025-10-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Notably:

* Does not implement for the streaming case, as it seems like this would require more invasive changes.
* The duration is set _after_ `value_turn()` creates the Turn so that we don't have to tamper with the `value_turn()` methods provider-by-provider. Integrating into each `value_turn()` method might be the more sound thing to do, but would just result in much more code.

Filing from a fork as I'm having some git wonk at the moment. Opening as a draft because my brain is a bit fried at the moment—will revisit in the morning and then request review if all seems well. :)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* Turns now have a `@duration` slot. The slot is `NA` for user turns and a numeric giving the total time to complete the request for assistant turns (@simonpcouch, #798).
 * New `chat_openai_responses()` to use the new OpenAI responses API (#365).
 * `parallel_chat_structured()` now returns a tibble, since this does a better job of printing more complex data frames (#787).
 * `parallel_chat()` and friends now have a more permissive attitude to errors. By default, they will now return when hitting the first error (rather than erroring), and you can control this behaviour with the `on_error` argument. Or if you interrupt the job, it will finish up current requests and then return all the work done so far. The main downside of this work is that the output of `parallel_chat()` is more complex: it is now a mix of `Chat` objects, error objects, and `NULL` (#628).


R/httr2.R
```diff
@@ -22,17 +22,13 @@ chat_perform <- function(
 
   switch(
     mode,
-    "value" = chat_perform_value(provider, req),
+    "value" = req_perform(req),
     "stream" = chat_perform_stream(provider, req),
-    "async-value" = chat_perform_async_value(provider, req),
+    "async-value" = req_perform_promise(req),
```
simonpcouch, line 27: The removed helpers were invoked only once each. 

This function, `chat_perform()`, is only called twice, both times in `chat.R`. In both of those usage, we now just call `resp_body_json()` on the output of `req_perform()` or `req_perform_promise()` and, now, `resp_timing()` additionally.

R/httr2.R
```diff
@@ -50,10 +46,6 @@ on_load(
   })
 )
 
-chat_perform_async_value <- function(provider, req) {
-  promises::then(req_perform_promise(req), resp_body_json)
-}
-
 on_load(
   chat_perform_async_stream <- coro::async_generator(function(provider, req) {
     resp <- req_perform_connection(req, blocking = FALSE)
```
simonpcouch, line 51: An implementation for streaming responses could pass along this `resp` (or just extract `resp_timing["total"]` and pass only that along), but I don't have as much use for streamed responses.

PR #802, opened 2025-10-22
Improve provider hashing for `batch_chat_*()`

hadley:
Rather than hashing the complete provider object, we should just hash the most critical components. This trades a little safety for convenience; I don't think there are likely to be too many cases where you re-call `batch_chat()` with a subtly different provider object and this makes it more likely that you can retrieve results stored with a different version of ellmer. I've also added an escape hatch so you can use chats saved by ellmer 0.3.0 in ellmer 0.4.0

Extracted out from #799.

@simonpcouch I think the implementation is straightforward so doesn't need a close inspection, but does the reasoning and documentation make sense to you?


tests/testthat/test-batch-chat.R
```diff
@@ -43,6 +43,13 @@ test_that("errors if chat/provider/prompts don't match previous run", {
   )
 })
 
+test_that("can override hash check", {
+  chat <- chat_openai_test(system_prompt = "Be cool")
```
simonpcouch, line 47: 😎

R/batch-chat.R
```diff
@@ -332,25 +354,31 @@ BatchJob <- R6::R6Class(
...
+            i = "Or set {.code ignore_hash = TRUE} to ignore this check?"
+          ),
+          call = call
+        )
+      }
     }
   )
 )
 
 provider_hash <- function(x) {
```
simonpcouch, line 376: I assumed by the name of this function that it computed a hash on a subset of the provider rather than returned a subset of the provider

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # ellmer (development version)
 
+* `batch_*()` now only hash the `name`, `model`, and `base_url` of the provider. This should provide some protection from accidentally reusing the same `.json` file with different providers, while still allowing you to use the same batch file across ellmer versions.
```
simonpcouch:
```suggestion
* `batch_*()` no longer hashes properties of the provider besides the `name`, `model`, and `base_url`. This should provide some protection from accidentally reusing the same `.json` file with different providers, while still allowing you to use the same batch file across ellmer versions.
```

The language raised an eyebrow for me at first, as the hash is the provider info + prompts + turns—this read to me like the hash was now _just_ (a subset of) the provider info.

simonpcouch (approved):
Yes, this is reasonable!

PR #806, opened 2025-10-23
Deprecationapalooza

### LINKED ISSUE #805: Advance deprecations

hadley:
Remove all functions/arguments/methods deprecated in ellmer 0.1.0 and 0.2.0. This is a bit more aggressive than our usual policy but ellmer isn't 1.0.0 yet, so this is ok.

Fixes #805


hadley:
Thanks! I grepped for `deprecate_warn` so I'm pretty confident that I got it all. (Also I had claude generate the bulleted list for the news so I think that's an adequate second pair of eyes for this task).

PR #808, opened 2025-10-27
Deprecate `virtual_key` argument in `chat_portkey()`

### LINKED ISSUE #786: Switch from Virtual Keys to Model Catalog for Portkey
The upgrade is described [here](https://portkey.ai/docs/support/upgrade-to-model-catalog), it comes in life on 20th October 2025.

It is not breaking though, but as I get it streamlines portkey workflow. This will entail somehow deprecation of `virtual_key` param and switching to new one (`model_catalog`).

FYI @shajoezhu

maciekbanas:
Fixes #786

### DIFF
--- a/R/provider-portkey.R
+++ b/R/provider-portkey.R
@@ -79,7 +79,7 @@ chat_portkey <- function(
 
 chat_portkey_test <- function(
   ...,
-  model = "@open-ai-virtual-7f0dcd/gpt-4o-mini",
+  model = "@open-ai-virtual-7f0dcd/gpt-4.1-nano",
   params = NULL,
   echo = "none"
 ) {
--- a/tests/testthat/test-provider-portkey.R
+++ b/tests/testthat/test-provider-portkey.R
@@ -26,9 +26,8 @@ test_that("can extract data", {
 })
 
 test_that("can use images", {
-  chat_fun <- \(...) {
-    chat_portkey_test(model = "@open-ai-virtual-7f0dcd/gpt-4.1-mini", ...)
-  }
+  # Needs mini to get shape correct
+  chat_fun <- \(...) chat_portkey_test(model = "gpt-4.1-mini", ...)
 
   test_images_inline(chat_fun)
   test_images_remote(chat_fun)


maciekbanas:
@shajoezhu would you like to test if the new solution works for your examples?

R/provider-portkey.R
```diff
@@ -36,7 +35,13 @@ chat_portkey <- function(
 ) {
   model <- set_default(model, "gpt-4o")
   echo <- check_echo(echo)
-
+  if (lifecycle::is_present(virtual_key)) {
+    lifecycle::deprecate_warn(
```
hadley, line 44: I wonder if it's too aggressive to immediately deprecate? Is there some other way we could tell if the user has opted in to the model catalog?

shajoezhu:
> @shajoezhu would you like to test if the new solution works for your examples?

will do. thank you!

R/provider-portkey.R
maciekbanas: I think you may be right. I will get back with a more gentle solution.

maciekbanas: @hadley I introduced new `model_provider` parameter next to `virtual_key` - though, it is also optional.

shajoezhu:
i think it is nice to have the deprecation option for the virtual_key, it seems that free account has not fully moved away from the virtual key?

maciekbanas:
@shajoezhu you mean the public free account? Yes, they still use virtual keys, I am a bit confused. Still, in my solution, you can still pass `virtual_key` to `chat_portkey()`.

R/provider-portkey.R
maciekbanas: It works now this way - if there is no `model_provider` (passed as empty character `""`) nor `virtual_key`, `chat_portkey()` still chats, if the given `model`, e.g. "gpt" is available in user's Portkey AI. It seems like defining `model_provider` is not necessary (but it probably depends on configs/settings used by different hosts).

If user passes proper `model_provider`, that he has got in his Model Catalog, `chat_portkey()` will work, if wrong `model_provider` will be passed (non-existing in Model Catalog), an error occurs. Technically, in Portkey, model provider can be passed through:
* body 
```
curl https://api.portkey.ai/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "x-portkey-api-key: $PORTKEY_API_KEY" \
  -d '{
    "model": "@openai-prod/gpt-4o",
    "messages": [{"role": "user", "content": "Hello!"}]
  }'
```
* or headers:
```
curl https://api.portkey.ai/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "x-portkey-api-key: $PORTKEY_API_KEY" \
  -H "x-portkey-provider: @openai-prod" \
  -d '{
    "model": "gpt-4o",
    "messages": [{"role": "user", "content": "Hello!"}]
  }
```
I applied second solution to `ellmer` (headers).

FYI @shajoezhu @hadley

shajoezhu:
> @shajoezhu you mean the public free account? Yes, they still use virtual keys, I am a bit confused. Still, in my solution, you can still pass `virtual_key` to `chat_portkey()`.

hi @maciekbanas , yes, i agree with your solution. there is no problem here

hadley:
@maciekbanas I took a slightly different direction because my reading on the portkey docs suggests that they want you to write `@provider/model` rather than supplying a `model_provider` separately. Take a look and let me know what you think.

maciekbanas:
@hadley I think your solution is more clear - we need to specify a provider together with the model, as recommended by Portkey. If not, client tries to reach for a virtual key. I like it. Wonder what is @shajoezhu 's opinion. Joe, can you test it on your use cases?

hadley:
I want to get this into the next ellmer release, so I'm going to merge shortly. But if @shajoezhu has any additional feedback I'm happy to take it into account in the next release.

shajoezhu:
Thanks a lot @hadley , I have nothing to add. The new changes are neat, and works perfectly. 

> @hadley I think your solution is more clear - we need to specify a provider together with the model, as recommended by Portkey. If not, client tries to reach for a virtual key. I like it. Wonder what is @shajoezhu 's opinion. Joe, can you test it on your use cases?

@maciekbanas , I had not realized we needed to update the model as "@provider/model", so took me some time to find out why it was not working for me. But all is good now. 

Thanks again!

Issue #804, opened 2025-10-23
`Chat`, not the `Provider` implementation, should call `value_tokens()`


cpsievert:
See this discussion https://github.com/tidyverse/ellmer/pull/784#discussion_r2452358120

hadley:
Now happens in `Chat$add_turn()`

PR #812, opened 2025-10-29
Refactor cost computation

hadley:
Now we compute the cost once, when we create the turn. The makes a lot of things simpler, e.g. we no longer need to track the variant in `the$tokens`, and unit tests get simpler. I also introduce `Chat$get_cost_details()` (ported from #810) to make it easier to inspect the raw details of a chat.

Includes a fix for #804 by logging as part of `Chat$add_turn()`. This breaks #743 in a new way: for now batch chat usage isn't logged at all. But I think it makes sense to fix that in another PR.

I've also added `map_tokens()` helper to hide the awkward use of `vapply()` in a few places (and to make the output more consistent)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* New `Chat$get_cost_details()` to get turn-by-token token usage + costs (#812).
 * Updated pricing data (#790).
 * The following deprecated functions/arguments/methods have now been removed:
   * `Chat$extract_data()` -> `chat$chat_structured()` (0.2.0)


R/chat.R
```diff
@@ -75,6 +75,13 @@ Chat <- R6::R6Class(
       check_turn(user)
       check_turn(system)
 
+      tokens_log(
+        private$provider,
+        # TODO: store better representation in Turn object
+        exec(tokens, !!!as.list(system@tokens)),
+        system@cost
```
cpsievert: `system` here really means `assistant`, right? Worth fixing that naming now?

R/turns.R
```diff
@@ -54,14 +57,19 @@ Turn <- new_class(
...
     )
   ),
   constructor = function(
     role,
     contents = list(),
     json = list(),
     tokens = c(0, 0, 0),
-    duration = NA_real_
+    duration = NA_real_,
+    cost = NA_real_
```
cpsievert: I'd prefer `tokens` and `cost` to be next to one another

R/chat.R
hadley: Oh yeah that confused me too. Might as well fix now.

PR #848, opened 2025-11-07
Fix the tracer within ellmer so that it is not dynamic during testing

schloerke:
Instead... reset ellmer's otel tracer when calling `otelsdk::with_otel_record()` during testing. 

Motivation: If someone is testing their R package that uses the ellmer package, they should not receive ellmer's otel spans (or any other package's otel spans). They should only receive their package's otel spans.

-------------------------

Until `{otel}`/`{otelsdk}` can handle turning packages on/off on demand, we should be cautious in how we enable tracing. For now, do not trace during external testing.

### DIFF
--- a/tests/testthat/helper-otel.R
+++ b/tests/testthat/helper-otel.R
@@ -1,0 +1,11 @@
+with_ellmer_otel_record <- function(...) {
+  stopifnot(is_testing())
+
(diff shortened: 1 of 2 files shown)


schloerke:
(untested)

PR #870, opened 2025-11-18
✨ forward 'thoughtSignature' with Google Gemini

### LINKED ISSUE #869: Check Gemini thinking signatures
https://ai.google.dev/gemini-api/docs/thought-signatures

simonpcouch:
Closes #869.

``` r
library(ellmer)

tool_rnorm <- tool(
  function() rnorm(1),
  description = "Draw a number from a standard normal distribution"
)

gemini_3 <- chat_google_gemini(
  system_prompt = "Be terse.",
  model = "gemini-3-pro-preview"
)
gemini_3$register_tool(tool_rnorm)

gemini_3$chat("Draw a number from a standorm normal.")
#> ◯ [tool call] tool_001()
#> ● #> 0.8896
#> 0.8896
```

<sup>Created on 2025-11-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/provider-google.R
+++ b/R/provider-google.R
@@ -327,10 +327,9 @@ method(value_turn, ProviderGoogleGemini) <- function(
         ContentText(content$text)
       }
     } else if (has_name(content, "functionCall")) {
-      extra <- list()
-      if (has_name(content, "thoughtSignature")) {
-        extra$thoughtSignature <- content$thoughtSignature
-      }
+      extra <- compact(list(
+        thoughtSignature = content$thoughtSignature
+      ))
       ContentToolRequest(
         content$functionCall$name,
         content$functionCall$name,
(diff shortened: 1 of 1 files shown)


hadley:
Also did we need some adjustment to params? I noticed you didn’t use it in the example.

simonpcouch:
You got it, thank you!

> Also did we need some adjustment to params? I noticed you didn’t use it in the example.

This references using `params(reasoning_effort)` rather than `api_args(...(thinkingLevel))`?

```r
gemini_3 <- chat_google_gemini(
  model = "gemini-3-pro-preview",
  api_args = list(
    generationConfig = list(thinkingConfig = list(thinkingLevel = "low"))
  )
)
```

There's a default `thinkingLevel`, so setting that argument isn't required. The above is the same syntax as was used for Gemini 2.5 Pro, so I figured there might have been some other reason yall didn't incorporate that argument into `params(reasoning_effort)`. I do agree that setting that `params` entry would be easier, but since this isn't a new API argument I left it for another PR.

PR #890, opened 2025-11-26
resolve OpenRouter tool calling failures

### LINKED ISSUE #889: intermittent failures tool calling with OpenRouter
On OpenRouter, even models like Claude Sonnet seem to have trouble consistently passing arguments that are compliant with the tool description. Performance of models seems to degrade on OpenRouter compared to the usual provider (in this case, Anthropic, but I observed this with many other model families). I also tested setting the provider within OpenRouter to the same API that the native provider ultimately hits, and still saw that degradation. So, I think there's an actual issue with the implementation of `chat_openrouter()`.

It appears that, when conversation history is sent back to OpenRouter, `jsonlite::toJSON()` serializes tool call arguments without `auto_unbox = TRUE`, wrapping scalar values in arrays, e.g. `"path":["R/file.R"]` instead of `"path":"R/file.R"`. The model sees this format and intermittently mimics it in subsequent responses, producing malformed arguments like `"[\"file\"]"`.

Here's a reprex--I had hoped to make it more minimal, but it seems to require two different, relatively complex tools, called in sequence. Even then, this only failed after 5 successes. (Note that it needs to be ran with `wd = "."`.) Note the `limit = "[100]", case_sensitive = "[false]"` rather than `"100"` and `FALSE`.

``` r
library(ellmer)

# for an example tool that has some non-trivial arguments
library(btw)

client <- chat_openrouter(model = "anthropic/claude-haiku-4.5")
client$set_system_prompt("Be very terse.")
client$register_tool(btw_tool_files_list_files)
client$register_tool(btw_tool_files_code_search)

client$chat("Is there a file in the R directory with util in the name?")
#> I'll search for files in the R directory with "util" in the
#> name.
#> ◯ [tool call] btw_tool_files_list_files(path = "R", type =
#> "file", regexp = "util", ...)
#> ● #> | path | type | size | modification_time |
#>   #> |------|------|------|-------------------|
#>   #> | R/utils.R | file | 82 | 2025-11-24 09:27:56 |
#> Yes, there is a file in the R directory with "util" in the 
#> name: **`R/utils.R`**. It's 82 bytes in size and was last 
#> modified on 2025-11-24 at 09:27:56.
#> I'll search for "util" in the contents of all .R files in 
#> the project.
client$chat("How about util in the contents of any .R files?")
#> ◯ [tool call] btw_tool_files_code_search(term = "util",
#> limit = 100L, case_sensitive = FALSE, ...)
#> ✔ Indexing files in /Users/simoncouch/Documents/rrr/helperb…
#> ● #> [{"filename":"inst/bundles/openrouter-degrade/ass…
#> The search returned mostly matches from bundled JavaScript 
#> files. Let me refine the search to focus specifically on .R
#> files:
#> ◯ [tool call] btw_tool_files_code_search(term = "util",
#> limit = "[100]", case_sensitive = "[false]", ...)
#> ■ #> Error: `limit` must be a whole number, not the string
#> "[…
#> Let me fix that:
#> ◯ [tool call] btw_tool_files_code_search(term = "util",
#> limit = 100L, case_sensitive = FALSE, ...)
#> ✔ Indexing files in /Users/simoncouch/Documents/rrr/helperb…
#> ● #> [{"filename":"inst/bundles/openrouter-degrade/ass…
#> The search is picking up bundled files. Let me list the R 
#> directory to see what .R files are there:
#> ◯ [tool call] btw_tool_files_list_files(path = "R", type =
#> "file", regexp = "\\.R$", ...)
#> ● #> | path | type | size | modification_time |
#>   #> |------|------|------|-------------------|
#>   #> | R/helper-client.R | file | 458 | 2025-11-24 08:49:36
#> |
#>   #> | R/helper-dataset.R | file | 548 | 2025-11-24
#> 09:09:27 |
#>   #> | R/helper-scorer.R | file | 3.01K | 2025-11-24
#> 20:21:48 |
#>   #> …
#> Good! Now let me search specifically in those R files for 
#> "util":
#> ◯ [tool call] btw_tool_files_code_search(term =
#> "[\"util\"]", limit = "[100]", ...)
#> ■ #> Error: `limit` must be a whole number, not the string
#> "[…
#> Let me fix that parameter issue:
#> ◯ [tool call] btw_tool_files_code_search(term = "util",
#> limit = 100L, case_sensitive = FALSE, ...)
```

<sup>Created on 2025-11-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

First noted in https://github.com/simonpcouch/helperbench/issues/1.

simonpcouch:
Closes #889.

``` r
  library(ellmer)

# for an example tool that has some non-trivial arguments
library(btw)

client <- chat_openrouter(model = "anthropic/claude-haiku-4.5")
client$set_system_prompt("Be very terse.")
client$register_tool(btw_tool_files_list_files)
client$register_tool(btw_tool_files_code_search)

client$chat("Is there a file in the R directory with util in the name?")
#> ◯ [tool call] btw_tool_files_list_files(path = "R", type = "file", regexp =
#> "util", ...)
#> ● #> | path | type | size | modification_time |
#>   #> |------|------|------|-------------------|
#>   #> | R/chat-utils.R | file | 336 | 2025-11-26 10:14:30 |
#>   #> | R/utils-S7.R | file | 4.71K | 2025-11-26 10:14:30 |
#>   #> | R/utils-auth.R | file | 2.13K | 2025-11-26 10:14:30 |
#>   #> …
#> Yes, there are several files with "util" in the name in the R directory:
#> 
#> - `chat-utils.R`
#> - `utils-S7.R`
#> - `utils-auth.R`
#> - `utils-callbacks.R`
#> - `utils-cat.R`
#> - `utils-coro.R`
#> - `utils-merge.R`
#> - `utils-prettytime.R`
#> - `utils.R`
client$chat("How about util in the contents of any .R files?")
#> ◯ [tool call] btw_tool_files_code_search(term = "util", limit = 100L,
#> case_sensitive = FALSE, ...)
#> ℹ Indexing files in '/Users/simoncouch/Documents/rrr/helperbench/inst/resources…
#> ✔ Indexing files in '/Users/simoncouch/Documents/rrr/helperbench/inst/resources…
#> 
#> ● #> [{"filename":"piny-loon_reprex.spin.R","size":482,"last_modified":"2025-1…
#> Yes, "util" appears in many .R files. Key occurrences include:
#> 
#> - **R/utils-*.R files** (multiple files with "Utility functions" comments)
#> - **utils:: calls** throughout (e.g., `utils::packageVersion()`, 
#> `utils::capture.output()`, `utils::modifyList()`)
#> - **Comments** like "Utilities" and "Utility functions"
#> 
#> The search returned 33 matches across various R files, mostly referencing 
#> utility functions from R's `utils` package or comments describing utility 
#> functionality.
```

<sup>Created on 2025-11-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/tests/testthat/_snaps/provider-openrouter.md
+++ b/tests/testthat/_snaps/provider-openrouter.md
@@ -5,11 +5,11 @@
     Condition
       Error in `req_perform()`:
       ! HTTP 400 Bad Request.
-      * Expected temperature to be a number, received "hot"
+      i Invalid input: expected number, received string
     Code
       chat$chat("What is 1 + 1?", echo = TRUE)
     Condition
       Error in `req_perform_connection()`:
       ! HTTP 400 Bad Request.
-      * Expected temperature to be a number, received "hot"
+      i Invalid input: expected number, received string
 


R/provider-openrouter.R
```diff
@@ -173,3 +173,16 @@ method(as_json, list(ProviderOpenRouter, ContentText)) <- function(
     list(type = "text", text = x@text)
   }
 }
+
+method(as_json, list(ProviderOpenRouter, ContentToolRequest)) <- function(
+  provider,
+  x,
+  ...
+) {
+  json_args <- jsonlite::toJSON(x@arguments, auto_unbox = TRUE)
```
simonpcouch: This is exactly the same as the previous method OpenRouter dispatched to from [OpenAICompatible](https://github.com/tidyverse/ellmer/blob/5ab7aca525062181711c024fba221cf5454d0812/R/provider-openai-compatible.R#L403-L414), but with `auto_unbox = TRUE`. It may be possible that this fix ought to just live there, but I'm not sure what providers yall tend to get mileage with there.

hadley: Yeah definitely move it there. It’s surprising that this hasn’t caused more problems but some of these systems are surprisingly robust.

hadley:
Hmmmm, actually would you mind adding a `to_json()` helper and then using it every where. It looks like this might also affect a couple of other providers.

PR #900, opened 2025-12-05
✨ Fix optional types to return NULL when data unavailable (Claude provider) ✨

### LINKED ISSUE #899: Structured output - Claude's handling of optional `required=FALSE` types always returns a string if value is unknown
Very cool how this is currently spoofed with tool calls :)  Just found this when working on some course materials, though looks like it'll probably will be solved by #866.

```r
library(ellmer)

colors <- type_string(required = FALSE)

chat <- chat_github()
chat$chat_structured("What's my favourite colour?", "My favourite food is cake", type = colors)
#> Using model = "gpt-4.1".
#> NULL

chat <- chat_anthropic()
chat$chat_structured("What's my favourite colour?", "My favourite food is cake", type = colors)
#> Using model = "claude-sonnet-4-5-20250929".
#> [1] "I don't have information about your favourite colour. You mentioned that your favourite food is cake, but you haven't told me what your favourite colour is. Would you like to tell me your favourite colour?"
```

It's a bit trickier when the type is, say, an integer not a string as it'll still return the string so breaks the type "guarantee" now.

``` r
library(ellmer)

fave_num <- type_integer(required = FALSE)

chat <- chat_anthropic()
#> Using model = "claude-sonnet-4-5-20250929".
chat$chat_structured("What's my favourite number?", "My favourite food is cake", type = fave_num)
#> [1] "<UNKNOWN>"
```

thisisnic:
Fixes #899 - optional types (`required = FALSE`) in structured data extraction would return explanatory text instead of `NULL` when the requested information was not available in the conversation, when using Claude.

It's >90% Claude but with lots of updated instructions from me to get to this version, and some tweaks.

### DIFF
--- a/tests/testthat/_vcr/anthropic-optional-array.yml
+++ b/tests/testthat/_vcr/anthropic-optional-array.yml
@@ -4,38 +4,39 @@ http_interactions:
     uri: https://api.anthropic.com/v1/messages
     body:
       string: '{"model":"claude-sonnet-4-5-20250929","messages":[{"role":"user","content":[{"type":"text","text":"\n    Apples
-        are tasty.\n    Oranges are delicious.\n    Bananas are yellow and nutritious.\n  ","cache_control":{"type":"ephemeral","ttl":"5m"}}]}],"stream":false,"tools":[{"name":"_structured_tool_call","description":"Extract
+        are tasty.\n    Oranges are orange in colour.\n    Bananas are nutritious,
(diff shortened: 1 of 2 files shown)


thisisnic:
I ran the tests for the Claude provider specifically locally, and they passed.  This PR fails on CI when re-rendering vignettes which require both Claude and OpenAI as I don't have an OpenAI API key so couldn't update the cassette locally.

thisisnic:
Converting to draft as it returns NULL in some cases where NA would be better

PR #904, opened 2025-12-13
Handle NULL service_tier in ProviderOpenAI

### LINKED ISSUE #903: get_token_cost() fails when API response lacks service_tier
## Problem

I'm using ellmer with my company's internal OpenAI endpoint.

This worked in ellmer v0.3.0:

```r
library(ellmer)
chat <- chat_openai(
  model = "bedrock-anthropic-claude-3.7-sonnet",
  base_url = "internal-endpoint",
)
chat$chat("What is the difference between a tibble and a data frame?")
```

In ellmer v0.4.0, I'd start to get a response, then:

```
Error in data.frame(provider = provider@name, model = provider@model,
  arguments imply differing number of rows: 1, 0
```

## Likely Cause

`get_token_cost()` is called with `variant = NULL` when the API response doesn't include [`service_tier`](https://github.com/tidyverse/ellmer/blob/d26b150989eba931fa86ad46845a82f0bf54bb13/R/provider-openai.R#L326). The `data.frame()` call then fails because `variant` has length 0 while other columns have length 1.

## Potential Fix

Default variant to `NULL` and set to `""` if so:

```r
get_token_cost <- function(provider, tokens, variant = NULL) {
  variant <- variant %||% ""
  # ...
}
```

trangdata:
Resolves #903.

Fixes errors when using `chat_openai()` with OpenAI endpoints that don't return `service_tier` in their responses. `value_turn()` now default `service_tier` to "default" when NULL.


trangdata:
Hi @gadenbuie any chance you could review this simple PR quick so we could use it for with Azure endpoint? 🙏🏽

hadley:
@trangdata sorry for missing this! I'll merge momentarily.

PR #906, opened 2025-12-16
feat: enable contents_record/replay to handle lists (#895)

### LINKED ISSUE #895: `contents_record()` and `contents_replay()` could operate on list of turns
```r
chat$get_turns() |> ellmer::contents_record()
Run `rlang::last_trace()` to see where the error occurred.
Error in `ellmer::contents_record()`:
! Cannot record or replay a list.
ℹ Only `ellmer::Content` or `ellmer::Turn` classes or subclasses are currently supported.
```

But it'd be very convenient to be able to round trip

```r
first <- chat$get_turns() |> contents_record()
chat2$set_turns(contents_replay(first))
```

hopessugar:
Closes #895.

This PR updates `contents_record()` and `contents_replay()` to automatically handle lists of turns/objects. This simplifies saving and loading chat history by removing the need for manual `lapply` calls, as requested.

### DIFF
--- a/tests/testthat/_snaps/content-replay.md
+++ b/tests/testthat/_snaps/content-replay.md
@@ -2,9 +2,8 @@
 
     Code
       contents_replay(bad_names)
-    Condition
-      Error in `contents_replay()`:
-      ! Expected the recorded object to be a list with at least names 'version', 'class', and 'props'.
+    Output
+      list()
     Code
       contents_replay(bad_version)
     Condition


gadenbuie:
Thanks, but unfortunately, the fix for this issue is going to be more complicated. We'd need to introduce an intermediate class to ensure the general type-safety of `contents_replay()` and `contents_record()`. I mentioned this in https://github.com/tidyverse/ellmer/issues/895#issuecomment-3597939316 but could have been clearer that we're not wanting to follow the approach initially mentioned in the issue. I kept that code in case it's useful to someone, but it's not intended to be incorporated into ellmer as such.

Issue #828, opened 2025-11-04
`stream_text()` -> `stream_content()`


hadley:
i.e. it should return a Content object with the delta

thisisnic:
Closing as this will be covered by #775

PR #909, opened 2025-12-19
Return `ContentThinking` objects from `stream_text()`

simonpcouch:
Addresses #837, and a subset of #828. Companion to https://github.com/posit-dev/shinychat/pull/167.

Changes `stream_text()` to return `Content` objects instead of strings, allowing downstream consumers to distinguish between content types during streaming. With the companion shinychat PR, results look like this:

https://github.com/user-attachments/assets/62e04149-8288-421d-8730-d6f784de6c92

I'm not sure if this is the opposite direction of the dev dependency that yall would usually introduce. With `main` shinychat and this PR, each thinking delta would otherwise get its own collapsible:

collapsible-per-chunk

So, I opted to make ellmer depend on the shinychat Remotes. If this is a no-go, we could add some version-dependent logic, but felt cleaner to start with this approach.

Here's the reprex I use in the video:

```r
# pak::pak("posit-dev/shinychat/pkg-r#167")
library(ellmer)
library(shinychat)

# Anthropic with thinking ------------------------------------------
chat_anthropic <- chat_claude(
  model = "claude-sonnet-4-5-20250929",
  params = params(reasoning_tokens = 1024)
)

clipr::write_clip(
  "Write a limerick about the tidyverse, returning only the limerick."
)
chat_app(chat_anthropic)

# OpenAI with reasoning ------------------------------------------
chat_openai <- chat_openai(
  model = "gpt-5.1",
  params = params(reasoning_effort = "high")
)

chat_app(chat_openai)

# Gemini with thinking ------------------------------------------
chat_gemini <- chat_google_gemini(
  model = "gemini-3-flash-preview",
  params = params(reasoning_tokens = 1024)
)

chat_app(chat_gemini)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # ellmer (development version)
 
-* `stream_text()` now returns `ContentThinking` objects for thinking/reasoning content from Anthropic, OpenAI, and Google Gemini, allowing downstream packages like shinychat to distinguish thinking from regular text during streaming (@simonpcouch, #909).
+* ellmer will now distinguish text content from thinking content while streaming, allowing downstream packages like shinychat to provide specific UI for thinking content (@simonpcouch, #909).
 * `chat_github()` now uses `chat_openai_compatible()` for improved compatibility, and `models_github()` now supports custom `base_url` configuration (@D-M4rk, #877).
 * `chat_ollama()` now contains a slot for `top_k` within the `params` argument (@frankiethull).
 


R/chat.R
```diff
@@ -606,11 +606,16 @@ Chat <- R6::R6Class(
...
         for (chunk in response) {
-          text <- stream_text(private$provider, chunk)
-          if (!is.null(text)) {
+          content <- stream_text(private$provider, chunk)
+          if (!is.null(content)) {
+            text <- switch(class(content)[1],
+              "ellmer::ContentThinking" = content@thinking,
+              "ellmer::ContentText" = content@text,
+              if (S7_inherits(content, Content)) format(content) else content
+            )
```
simonpcouch: I'm not sure if using `switch` on S7 objects is considered bad practice, but the control flow was gnarly otherwise😅

hadley: Maybe we should have a `stream_data()` that returns the object and then `stream_text()` would contain this code to extract out just the text?

hadley: Or `stream_content()`?

R/chat.R
```diff
@@ -774,6 +768,23 @@ Chat <- R6::R6Class(
   )
 )
 
+stream_content <- function(provider, event) {
+  result <- stream_text(provider, event)
```
hadley: This feels inside out to me? Why doesn't `stream_text()` call `stream_content()`?

simonpcouch: I had been wondering whether I was interpreting that incorrectly. :) Just switched those out.

R/provider.R
```diff
@@ -144,13 +144,30 @@ stream_parse <- new_generic(
...
+
+stream_text <- function(provider, event) {
+  content <- stream_content(provider, event)
+  if (is.null(content)) {
+    return(NULL)
+  }
+  content_text(content)
+}
+
+content_text <- function(content) {
```
hadley, line 163: Yes, you ate my feedback and left no crumbs.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* `stream_text()` now returns `ContentThinking` objects for thinking/reasoning content from Anthropic, OpenAI, and Google Gemini, allowing downstream packages like shinychat to distinguish thinking from regular text during streaming (@simonpcouch, #909).
```
hadley: This now needs an update, but otherwise looks good I think.

R/provider.R
simonpcouch: W PR

PR #917, opened 2026-01-22
Standardize `models_anthropic()` credentials handling to match other providers

jcrodriguez1989:
## Summary
Refactors `models_claude()` and `models_anthropic()` to use the same credential handling pattern as the other `models_*` functions (`models_openai()`, `models_google_gemini()`, etc.).

## Changes
- Added `credentials = NULL` parameter to function signature
- Changed `api_key` parameter default from `anthropic_key()` to `NULL`
- Used `as_credentials()` helper for consistent credential resolution


hadley:
This makes me wonder if it's worth extracting out `credentials_anthropic()` to reduce the duplication still further. (And similarly for for the other providers)

PR #926, opened 2026-02-14
Add Gemini batch processing support

### LINKED ISSUE #914: Feature Request: add Google Gemini support batch_chat()
`batch_chat()` works great for OpenAI models (haven't tried Claude). But I noticed that Google has enabled batch, not sure when. Here's [an example](https://ai.google.dev/gemini-api/docs/pricing) from it's documentation.

When you have resources, it would be great if batch_chat could be extended to Gemini models, given how well the latter are performing. 

No pressure, of course!

xmarquez:
## Summary

- Adds batch processing support for `chat_google_gemini()` via the [Gemini Batch API](https://ai.google.dev/gemini-api/docs/batch-api)
- Implements all 6 batch S7 methods directly on `ProviderGoogleGemini`: `has_batch_support`, `batch_submit`, `batch_poll`, `batch_status`, `batch_retrieve`, `batch_result_turn`
- Includes helper functions for JSONL preparation, file upload/download, and snake_case conversion
- Keeps polling when `BATCH_STATE_SUCCEEDED` but output file isn't available yet
- Updates `batch_chat()` documentation to include Gemini
- Adds pre-recorded fixture file for deterministic offline tests

Closes #914

## Test plan

- [x] 35 unit tests pass without credentials (helper functions, batch status, fixture-based batch_chat_text)
- [x] 2 integration tests skip gracefully without credentials
- [x] Live tested with both `gemini-2.5-flash` and `gemini-3-flash-preview`, using `batch_chat` and `batch_chat_structured` (4 scenarios, all pass)
- [x] Pre-recorded fixture at `tests/testthat/batch/state-capitals-gemini.json` returns correct state capitals
- [x] `devtools::test()`: 788 pass, 0 fail
- [x] `devtools::check()`: 0 errors, 0 warnings

### DIFF
--- a/R/batch-chat.R
+++ b/R/batch-chat.R
@@ -122,6 +122,9 @@ batch_chat_text <- function(
     wait = wait,
     ignore_hash = ignore_hash
   )
+  if (is.null(chats)) {
+    return()
+  }
   map_chr(chats, \(chat) {
     if (is.null(chat)) NA_character_ else chat$last_turn()@text
   })
--- a/R/batch-chat.R
+++ b/R/batch-chat.R
@@ -91,6 +91,9 @@ batch_chat <- function(chat, prompts, path, wait = TRUE, ignore_hash = FALSE) {
     ignore_hash = ignore_hash
   )
   job$step_until_done()
+  if (job$stage != "done") {
+    return(NULL)
+  }
 
   assistant_turns <- job$result_turns()
   map2(job$user_turns, assistant_turns, function(user, assistant) {
@@ -120,6 +123,9 @@ batch_chat_text <- function(
     wait = wait,
     ignore_hash = ignore_hash
   )
+  if (is.null(chats)) {
+    return(NULL)
+  }
   map_chr(chats, \(chat) {
     if (is.null(chat)) NA_character_ else chat$last_turn()@text
   })
@@ -152,6 +158,9 @@ batch_chat_structured <- function(
     ignore_hash = ignore_hash
   )
   job$step_until_done()
+  if (job$stage != "done") {
+    return(NULL)
+  }
   turns <- job$result_turns()
 
   multi_convert(
--- a/R/batch-chat.R
+++ b/R/batch-chat.R
@@ -2,10 +2,11 @@
 #'
 #' @description
 #' `batch_chat()` and `batch_chat_structured()` currently only work with
-#' [chat_openai()] and [chat_anthropic()]. They use the
-#' [OpenAI](https://platform.openai.com/docs/guides/batch) and
-#' [Anthropic](https://docs.claude.com/en/docs/build-with-claude/batch-processing)
-#' batch APIs which allow you to submit multiple requests simultaneously.
+#' [chat_openai()], [chat_anthropic()], and [chat_google_gemini()]. They use
+#' the [OpenAI](https://platform.openai.com/docs/guides/batch),
+#' [Anthropic](https://docs.claude.com/en/docs/build-with-claude/batch-processing),
+#' and [Google Gemini](https://ai.google.dev/gemini-api/docs/batch) batch APIs
+#' which allow you to submit multiple requests simultaneously.
 #' The results can take up to 24 hours to complete, but in return you pay 50%
 #' less than usual (but note that ellmer doesn't include this discount in
 #' its pricing metadata). If you want to get results back more quickly, or
--- a/R/batch-chat.R
+++ b/R/batch-chat.R
@@ -5,7 +5,7 @@
 #' [chat_openai()], [chat_anthropic()], and [chat_google_gemini()]. They use
 #' the [OpenAI](https://platform.openai.com/docs/guides/batch),
 #' [Anthropic](https://docs.claude.com/en/docs/build-with-claude/batch-processing),
-#' and [Google Gemini](https://ai.google.dev/gemini-api/docs/batch) batch APIs
+#' and [Google Gemini](https://ai.google.dev/gemini-api/docs/batch-api) batch APIs
 #' which allow you to submit multiple requests simultaneously.
 #' The results can take up to 24 hours to complete, but in return you pay 50%
 #' less than usual (but note that ellmer doesn't include this discount in
--- a/R/provider-google-upload.R
+++ b/R/provider-google-upload.R
@@ -43,6 +43,17 @@ google_upload <- function(
 
   mime_type <- mime_type %||% guess_mime_type(path)
 
+  status <- google_upload_file(
+    path = path,
+    base_url = base_url,
+    credentials = credentials,
+    mime_type = mime_type
+  )
+
+  ContentUploaded(uri = status$uri, mime_type = status$mimeType)
+}
+
+google_upload_file <- function(path, base_url, credentials, mime_type) {
   upload_url <- google_upload_init(
     path = path,
     base_url = base_url,
@@ -56,8 +67,7 @@ google_upload <- function(
     credentials = credentials
   )
   google_upload_wait(status, credentials)
-
-  ContentUploaded(uri = status$uri, mime_type = status$mimeType)
+  status
 }
 
 # https://ai.google.dev/api/files#method:-media.upload
@@ -133,20 +143,12 @@ gemini_upload_file <- function(
 ) {
   upload_base_url <- sub("/v[^/]+/?$", "/", provider@base_url)
 
-  upload_url <- google_upload_init(
+  google_upload_file(
     path = path,
     base_url = upload_base_url,
     credentials = provider@credentials,
     mime_type = mime_type
   )
-
-  status <- google_upload_send(
-    upload_url = upload_url,
-    path = path,
-    credentials = provider@credentials
-  )
-  google_upload_wait(status, provider@credentials)
-  status
 }
 
 gemini_download_file <- function(provider, name, path) {
--- a/R/provider-google-upload.R
+++ b/R/provider-google-upload.R
@@ -124,6 +124,39 @@ google_upload_wait <- function(status, credentials) {
   invisible()
 }
 
+# Batch file helpers -----------------------------------------------------------
+
+gemini_upload_file <- function(
+  provider,
+  path,
+  mime_type = "application/jsonl"
+) {
+  upload_base_url <- sub("/v[^/]+/?$", "/", provider@base_url)
+
+  upload_url <- google_upload_init(
+    path = path,
+    base_url = upload_base_url,
+    credentials = provider@credentials,
+    mime_type = mime_type
+  )
+
+  status <- google_upload_send(
+    upload_url = upload_url,
+    path = path,
+    credentials = provider@credentials
+  )
+  google_upload_wait(status, provider@credentials)
+  status
+}
+
+gemini_download_file <- function(provider, name, path) {
+  req <- base_request(provider)
+  req <- req_url_path_append(req, paste0(name, ":download"))
+  req <- req_url_query(req, alt = "media")
+  req_perform(req, path = path)
+  invisible(path)
+}
+
 # Helpers ----------------------------------------------------------------------
 
 guess_mime_type <- function(file_path, call = caller_env()) {
--- a/R/provider-google.R
+++ b/R/provider-google.R
@@ -888,7 +888,7 @@ google_location <- function(location) {
 
 # https://ai.google.dev/gemini-api/docs/batch
 method(has_batch_support, ProviderGoogleGemini) <- function(provider) {
-  TRUE
+  grepl("generativelanguage.googleapis.com", provider@base_url, fixed = TRUE)
 }
 
 method(batch_submit, ProviderGoogleGemini) <- function(
@@ -970,10 +970,24 @@ method(batch_status, ProviderGoogleGemini) <- function(provider, batch) {
     "BATCH_STATE_EXPIRED"
   )
 
+  is_terminal <- state %in% terminal_states
+
+  # Keep polling if succeeded but output file isn't available yet
+  if (state == "BATCH_STATE_SUCCEEDED") {
+    batch_resource <- batch$response %||% batch$metadata
+    responses_file <- batch_resource$output$responsesFile %||%
+      batch_resource$responsesFile %||%
+      metadata$output$responsesFile %||%
+      NULL
+    if (is.null(responses_file) || !nzchar(responses_file)) {
+      is_terminal <- FALSE
+    }
+  }
+
   n_processing <- max(pending, total - succeeded - failed, 0L)
 
   list(
-    working = !(state %in% terminal_states),
+    working = !is_terminal,
     n_processing = n_processing,
     n_succeeded = max(succeeded, 0L),
     n_failed = max(failed, 0L)
@@ -991,7 +1005,11 @@ method(batch_retrieve, ProviderGoogleGemini) <- function(provider, batch) {
     if (request_count <= 0L) {
       return(list(list(status_code = code, body = NULL)))
     }
-    return(replicate(request_count, list(status_code = code, body = NULL), simplify = FALSE))
+    return(replicate(
+      request_count,
+      list(status_code = code, body = NULL),
+      simplify = FALSE
+    ))
   }
 
   batch_resource <- batch$response %||% batch$metadata
@@ -1054,7 +1072,11 @@ gemini_prepare_batch_body <- function(body) {
     is_empty <- if (is.list(parts) && !is.null(names(parts))) {
       identical(parts$text, "") || is.null(parts$text)
     } else if (is.list(parts) && length(parts) > 0) {
-      all(vapply(parts, function(p) identical(p$text, "") || is.null(p$text), logical(1)))
+      all(vapply(
+        parts,
+        function(p) identical(p$text, "") || is.null(p$text),
+        logical(1)
+      ))
     } else {
       TRUE
     }
@@ -1075,7 +1097,9 @@ gemini_prepare_batch_body <- function(body) {
 
   # Rename response_schema -> response_json_schema and restore original schema
   gc <- body$generation_config
-  if (!is.null(gc) && (!is.null(gc$response_schema) || !is.null(saved_schema))) {
+  if (
+    !is.null(gc) && (!is.null(gc$response_schema) || !is.null(saved_schema))
+  ) {
     gc$response_json_schema <- saved_schema %||% gc$response_schema
     gc$response_schema <- NULL
     body$generation_config <- gc
@@ -1085,7 +1109,11 @@ gemini_prepare_batch_body <- function(body) {
 }
 
 #' @noRd
-gemini_upload_file <- function(provider, path, mime_type = "application/jsonl") {
+gemini_upload_file <- function(
+  provider,
+  path,
+  mime_type = "application/jsonl"
+) {
   upload_base_url <- sub("/v[^/]+/?$", "/", provider@base_url)
 
   upload_url <- google_upload_init(
@@ -1133,20 +1161,60 @@ gemini_extract_index <- function(x, default = NA_integer_) {
 #' @noRd
 gemini_json_fallback <- function(line) {
   index <- suppressWarnings(
-    as.integer(sub('.*"request_index"\\s*:\\s*([0-9]+).*', "\\1", line, perl = TRUE))
+    as.integer(sub(
+      '.*"request_index"\\s*:\\s*([0-9]+).*',
+      "\\1",
+      line,
+      perl = TRUE
+    ))
   )
 
   if (length(index) == 0L || is.na(index)) {
-    custom_id <- tryCatch({
-      m <- regmatches(line, regexpr('"custom_id"\\s*:\\s*"chat-[0-9]+"', line, perl = TRUE))
-      if (length(m) == 0L) NA_character_ else sub('.*"chat-([0-9]+)".*', "\\1", m)
-    }, error = function(e) NA_character_)
+    custom_id <- tryCatch(
+      {
+        m <- regmatches(
+          line,
+          regexpr('"custom_id"\\s*:\\s*"chat-[0-9]+"', line, perl = TRUE)
+        )
+        if (length(m) == 0L) {
+          NA_character_
+        } else {
+          sub('.*"chat-([0-9]+)".*', "\\1", m)
+        }
+      },
+      error = function(e) NA_character_
+    )
     index <- suppressWarnings(as.integer(custom_id))
   }
 
+  if (length(index) == 0L || is.na(index)) {
+    key_match <- tryCatch(
+      {
+        m <- regmatches(
+          line,
+          regexpr('"key"\\s*:\\s*"chat-[0-9]+"', line, perl = TRUE)
+        )
+        if (length(m) == 0L) {
+          NA_character_
+        } else {
+          sub('.*"chat-([0-9]+)".*', "\\1", m)
+        }
+      },
+      error = function(e) NA_character_
+    )
+    index <- suppressWarnings(as.integer(key_match))
+  }
+
   list(
-    metadata = if (length(index) == 0L || is.na(index)) list() else list(request_index = index),
-    status = list(code = 500L, message = "Failed to parse Gemini batch output line")
+    metadata = if (length(index) == 0L || is.na(index)) {
+      list()
+    } else {
+      list(request_index = index)
+    },
+    status = list(
+      code = 500L,
+      message = "Failed to parse Gemini batch output line"
+    )
   )
 }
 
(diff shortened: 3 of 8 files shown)


R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
 google_location <- function(location) {
   if (location == "global") "" else paste0(location, "-")
 }
+
+# Batched requests -------------------------------------------------------------
+
+# https://ai.google.dev/gemini-api/docs/batch
```
hadley: That doc link doesn't work. Hallucination?

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+method(has_batch_support, ProviderGoogleGemini) <- function(provider) {
+  grepl("generativelanguage.googleapis.com", provider@base_url, fixed = TRUE)
```
hadley: I think this could probably just be `TRUE` — folks may be using various enterprise proxy things and you still want batched requests to work there.

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+      request = gemini_prepare_batch_body(body)
+    )
+  })
+
+  json_lines <- map_chr(requests, to_json)
+  writeLines(json_lines, path)
+
+  uploaded <- gemini_upload_file(provider, path)
+  if (is.null(uploaded$name) || !nzchar(uploaded$name)) {
+    cli::cli_abort("Gemini upload did not return a file resource name.")
```
hadley: Can we either make this a bit more actionable for the user or mark it as `.internal = TRUE`?

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+  }
+
+  terminal_states <- c(
+    "BATCH_STATE_SUCCEEDED",
+    "BATCH_STATE_FAILED",
+    "BATCH_STATE_CANCELLED",
+    "BATCH_STATE_EXPIRED"
+  )
+
+  is_terminal <- state %in% terminal_states
```
hadley: I'd call this `is_done`

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+
+  # Keep polling if succeeded but output file isn't available yet
+  if (state == "BATCH_STATE_SUCCEEDED") {
+    batch_resource <- batch$response %||% batch$metadata
+    responses_file <- batch_resource$output$responsesFile %||%
+      batch_resource$responsesFile %||%
+      metadata$output$responsesFile %||%
+      NULL
+    if (is.null(responses_file) || !nzchar(responses_file)) {
```
hadley:
```suggestion
     ""
    if (!nzchar(responses_file)) {
```

I think that's a little simpler?

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+  response <- batch$response %||% list()
+  stats <- metadata$batchStats %||% response$batchStats %||% list()
+  request_count <- as.integer(stats$requestCount %||% 0L)
+
+  if (!is.null(batch$error)) {
+    code <- as.integer(batch$error$code %||% 500L)
+    if (request_count <= 0L) {
+      return(list(list(status_code = code, body = NULL)))
+    }
+    return(replicate(
```
hadley: Should use `rep()` not `replicate()` here

hadley: I think you could eliminate the branch above with `return(rep(list(status_code = code, body = NULL), min(0, request_count))`

But it might be safer to just error if `requiestCount` is <0?

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+      request_count,
+      list(status_code = code, body = NULL),
+      simplify = FALSE
+    ))
+  }
+
+  batch_resource <- batch$response %||% batch$metadata
+  responses_file <- batch_resource$output$responsesFile %||%
```
hadley: Hmmm, might be worth digging into why we look in so many places here?

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+  }
+
+  # Save user-defined schema before snake_case conversion so property names
+  # like "firstName" are not mangled to "first_name"
+  gc_pre <- body$generationConfig %||% body$generation_config
+  saved_schema <- if (!is.null(gc_pre)) {
+    gc_pre$responseSchema %||% gc_pre$response_schema
+  }
+
+  body <- gemini_to_snake_case(body)
```
hadley, line 1074: Are you sure this is necessary? Google APIs often seem to take both snake and camel case.

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+  ) {
+    gc$response_json_schema <- saved_schema %||% gc$response_schema
+    gc$response_schema <- NULL
+    body$generation_config <- gc
+  }
+
+  body
+}
+
+#' @noRd
```
hadley: No need for `@noRd` here and below

hadley: Should this and `gemini_download_file()`  live in `provider-google-upload.R`?

hadley: And is there a reason you can't use `google_upload()` here?

R/provider-google.R
```diff
@@ -872,3 +872,370 @@ models_google <- function(
...
+    NULL
+
+  if (is.null(responses_file) || !nzchar(responses_file)) {
+    cli::cli_abort("Gemini batch completed but no output file was returned.")
+  }
+
+  path_output <- withr::local_tempfile(fileext = ".jsonl")
+  gemini_download_file(provider, responses_file, path_output)
+
+  parsed <- read_ndjson(path_output, fallback = gemini_json_fallback)
```
hadley: Why the fallback here? Claude might have copied from OpenAI which seems to be flaky. Do you have evidence that gemini is similarly problematic?

hadley (commented):
Thanks for working on this! This is a great first step but I have lots of small questions/requests 😄

R/provider-google.R
xmarquez: Yes, sorry - it should have been <https://ai.google.dev/gemini-api/docs/batch-api>

xmarquez:
Thank you @hadley I'll work on this and get back to you this week. I've made a couple of replies but I will submit a revised PR shortly.

R/provider-google.R
xmarquez: The problem is that both `chat_google_gemini()` and `chat_google_vertex()` currently create the same ProviderGoogleGemini class in R/provider-google.R:43 and R/provider-google.R:94. So setting this to TRUE would also indicate batch support for Vertex, which I can't test using the live API (no access); this was an attempt to not advertise batch support on Vertex. But if you think it would be ok, I'll set it to `TRUE`.

R/provider-google.R
xmarquez: Claude tested with camelCase and the batch JSONL parser silently ignored the fields — it seems to require protobuf-style snake_case names, unlike the REST API which accepts both. The [batch API docs](https://ai.google.dev/gemini-api/docs/batch-api) also use snake_case in all their JSONL examples. So this seems to be necessary.

R/provider-google.R
xmarquez: (Have set it to `TRUE` in the revised PR)

R/provider-google.R
xmarquez: Have marked it `.internal = TRUE`

R/provider-google.R
xmarquez: Done - called it `is_done`.

R/provider-google.R
xmarquez: Yes thank you - have done that now.

R/provider-google.R
xmarquez: Used rep now. But I left the rest as is; ` min(0, request_count)` seems like it would always return 0. We could do `max(0, request_count)` or error if `requestCount ` is less than 0; let me know what you prefer!

R/provider-google.R
xmarquez: Ok, I've tested live with different models (gemini-2.5-flash and gemini-3-flash-preview) and both `batch_chat_text` and `batch_chat_structured` and it seems this was a survival from the fact that the file appears at both `batch$metadata$output$responsesFile` and and `batch$response$responsesFile`. But it seems that the final response is always consistently at `response$responsesFile`. Simplified to single path, which works in all the live tests I've run.

R/provider-google.R
xmarquez: Removed `@noRd` and moved `gemini_upload_file()` and `gemini_download_file()` to `provider-google-upload.R`. Re `google_upload()`: it returns a `ContentUploaded` object designed for inline content in chat messages. For batch, we need the raw file metadata, specifically `file.name` to pass as `inputConfig.fileName` in the batch request. So we use the lower-level `google_upload_init()`/`google_upload_send()`/`google_upload_wait()` directly. Happy to switch to `google_upload() `if there's a way to extract the file name from it that I'm missing.

R/provider-google.R
xmarquez: Removed the fallback. I tested with `gemini-2.5-flash` and `gemini-3-flash-preview` both in `batch_chat_text` and `batch_chat_structured`  and the JSONL is always well-formed, so it seems fine to remove. (I did run into an issue in a different place with the batch output from anthropic/claude, but I don't have a reprex; it concerned tool output in the json. So that's perhaps where this pattern might come from).

xmarquez:
Have now submitted a revised PR - hope I have done it correctly, I think it addresses all the comments. Thank you @hadley.

R/provider-google.R
hadley: Thanks for checking! Can you please add a brief summary as a comment?

R/provider-google.R
hadley: Yeah, it's probably fine. Thanks!

R/provider-google.R
hadley: Oops, yes I meant `max()`.

R/provider-google-upload.R
```diff
@@ -124,6 +124,39 @@ google_upload_wait <- function(status, credentials) {
   invisible()
 }
 
+# Batch file helpers -----------------------------------------------------------
+
+gemini_upload_file <- function(
```
hadley, line 139: I think you could reduce the duplication here by making this `google_upload_file()`, then `google_upload()` could call `google_upload_file()` then create the `ContentUploaded` object.

hadley (commented):
Getting very close now. Just a few last questions.

R/provider-google.R
xmarquez: Done now, using max(0L, request_count)

R/provider-google.R
xmarquez: Added a comment. Wrote another full test script to double-check the camelCase/snake_case problem, and it actually errors with HTTP 400 when it encounters camelCase (not silent), so have changed the other comment earlier in the file as well.

R/provider-google-upload.R
xmarquez: Done now - extracted `google_upload_file()`

xmarquez:
Kia ora @hadley - have addressed the final comments. Thank you for reviewing this! Also found another link that was problematic (https://ai.google.dev/gemini-api/docs/batch) and fixed it (to <https://ai.google.dev/gemini-api/docs/batch-api>).

xmarquez:
Hi @hadley - I've been doing some testing with a large-scale set of prompts using gemini batch and found a bug (after Codex did some extensive testing with my guidance). This concerned the ordering of the results - I was getting the batch results in the wrong order, despite the presence of the correct keys. Here's the report from Codex:

Findings

* Confirmed a real Gemini batch ordering bug in [provider-google.R (line 1096)]. The code was taking the fallback line number before checking the echoed key, so batch_retrieve() was effectively preserving raw file order. I verified that against your completed batch `ellmer-batch-acbf4e5ec13a252a6de1ac98b84bc70e-gemini-3.1-pro-preview-chat_gemini_extended.json`: the first saved response matches raw chat-85, not sorted chat-1. I fixed that and added regressions in [test-provider-google-batch.R (line 18)] and [test-provider-google-batch.R (line 193)].
* Confirmed a separate wait = FALSE bug in the public helpers. [batch-chat.R (line 94)], [batch-chat.R (line 126)], and [batch-chat.R (line 161)] would still fall through into result handling after a non-complete submit, which is why my live submit errored with “Expected 4, got 0.” I fixed that and covered it in [test-batch-chat.R (line 150)].
* I did not reproduce a deterministic “reasoning tokens always break Gemini structured JSON” bug on the current branch. I ran live parallel_chat_structured against chat_gemini_extended(model = "gemini-3.1-pro-preview", reasoning_tokens = 8192) with Populism-style prompts and parsing succeeded. I also inspected three completed project Gemini batch raw outputs: they all returned top-level key, none used code fences, none split JSON across multiple non-thought text parts, and the only bad row I found was a true per-request API error (Deadline expired before operation could complete.), not malformed JSON.

Focused verification passed with test-batch-chat.R and test-provider-google-batch.R. I also submitted two fresh 4-prompt Gemini 2.5 Flash batches during this session, but both were still pending after several minutes, so there isn’t a new completed live batch from this session yet.

Is it ok if I submit an amended PR to deal with the ordering issue?

hadley:
Thanks for the investigation. A PR would definitely be appreciated!

xmarquez:
Thanks @hadley I've submitted a revised PR with the two bug fixes. Let me know what you think!

R/batch-chat.R
```diff
@@ -90,6 +91,9 @@ batch_chat <- function(chat, prompts, path, wait = TRUE, ignore_hash = FALSE) {
     ignore_hash = ignore_hash
   )
   job$step_until_done()
+  if (job$stage != "done") {
+    return(NULL)
+  }
```
thisisnic: After rebasing, this change will end up being redundant - mind removing it?

tests/testthat/test-batch-chat.R
```diff
@@ -147,6 +147,32 @@ test_that("errors if wait = FALSE and not complete", {
...
+  ))
+  expect_null(batch_chat_structured(
+    chat,
+    prompts,
+    withr::local_tempfile(),
+    type = type_object(name = type_string()),
+    wait = FALSE
+  ))
+})
+
```
thisisnic: This chunk of code also becomes redundant after rebasing, sorry!

R/batch-chat.R
```diff
@@ -151,6 +158,9 @@ batch_chat_structured <- function(
...
```
thisisnic: Also becomes redundant after #979

R/batch-chat.R
```diff
@@ -119,6 +123,9 @@ batch_chat_text <- function(
     wait = wait,
     ignore_hash = ignore_hash
   )
+  if (is.null(chats)) {
+    return(NULL)
```
thisisnic:
```suggestion
    return()
```

No need to explicitly name `NULL` here as it's the default value.

R/provider-google.R
```diff
@@ -872,3 +872,275 @@ models_google <- function(
...
+  req <- req_url_path_append(req, batch$name)
+  resp <- req_perform(req)
+  resp_body_json(resp)
+}
+
+method(batch_status, ProviderGoogleGemini) <- function(provider, batch) {
+  metadata <- batch$metadata %||% list()
+  response <- batch$response %||% list()
+  state <- metadata$state %||% response$state %||% "BATCH_STATE_UNSPECIFIED"
+  stats <- metadata$batchStats %||% response$batchStats %||% list()
```
thisisnic: Are you sure that we need to check `response` for `state` and `batchStats`? 

In the fixture `state-capitals-gemini.json`, it only appears in `metadata`.  Same question for line 990 and `stats`.

R/provider-google.R
```diff
@@ -872,3 +872,275 @@ models_google <- function(
...
+      identical(parts$text, "") || is.null(parts$text)
+    } else if (is.list(parts) && length(parts) > 0) {
+      all(vapply(
+        parts,
+        function(p) identical(p$text, "") || is.null(p$text),
+        logical(1)
+      ))
+    } else {
+      TRUE
+    }
```
thisisnic: The `is_empty` check handles three shapes of parts (named list, unnamed list, empty). From what I can see, `chat_body()` for Google always produces parts as `list(text = "...")`. Are there other code paths that could produce a different shape? If not, could this be simplified?

R/provider-google.R
```diff
@@ -872,3 +872,275 @@ models_google <- function(
...
+    x$custom_id %||%
+    metadata$key %||%
+    metadata$custom_id %||%
+    ""
+  if (grepl("^chat-[0-9]+$", key)) {
+    return(as.integer(sub("^chat-([0-9]+)$", "\\1", key)))
+  }
+
+  as.integer(default)
+}
```
thisisnic, line 1091: I tested the batch API live across `gemini-2.5-flash`, `gemini-2.5-pro`, and `gemini-2.0-flash` (now deprecated) . The output JSONL always had key at the top level (e.g. `{"key": "chat-1", "response": {...}}`), including on error responses (`{"key": "chat-1", "error": {...}}`). I never saw `metadata`, `request_index`, `custom_id`, or `metadata$key` in any response, though could definitely be missing some testing scenarios! Are there scenarios where those fields would appear? 

The `custom_id` fields in particular look like they might be from the OpenAI convention. Could this potentially be simplified to just check `x$key` with the line-number fallback?

R/provider-google.R
```diff
@@ -872,3 +872,275 @@ models_google <- function(
...
+        result = list(status_code = 200L, body = x$response)
+      ))
+    }
+
+    status <- x$error %||% x$status %||% list()
+    code <- status$code %||% 500L
+    return(list(
+      index = index,
+      result = list(status_code = as.integer(code), body = NULL)
+    ))
```
thisisnic: I'm finding this a little hard to follow on a first read. I'm wondering if we could simplify it with sequential early returns  Something like:

```r
if (!is.null(x$response) && is.null(x$error) && is.null(x$status)) {                                                                                                              
    return(list(index = index, result = list(status_code = 200L, body = x$response)))
}                                                                                                                                                                                 
                                                                                                                                                                                      
if (!is.null(x$error) || !is.null(x$status)) {              
    code <- (x$error %||% x$status %||% list())$code %||% 500L                                                                                                                      
    return(list(index = index, result = list(status_code = as.integer(code), body = NULL)))                                                                                         
}
```

R/provider-google.R
thisisnic: After reviewing this PR in detail, I'm wondering if it might be worth restoring something in here to only return `TRUE` for the Gemini Developer API, as the Vertex API is drastically different, so won't we end up having users likely getting weird errors when it returns `TRUE` but then the API doesn't support the same mechanism?

R/provider-google.R
```diff
@@ -872,3 +872,275 @@ models_google <- function(
...
+  }
+
+  # Plain GenerateContentResponse lines (current file-mode output)
+  if (
+    !is.null(x$candidates) ||
+      !is.null(x$promptFeedback) ||
+      !is.null(x$usageMetadata)
+  ) {
+    return(list(index = index, result = list(status_code = 200L, body = x)))
+  }
```
thisisnic: Just wanted to check when this branch would get hit as I don't *think* this quite matches the developer API, though could be wrong

thisisnic (commented):
Thanks for making the PR!  I gave it a detailed run through, and it looks great at a high level, though I think there might be some bits where I'm either not ending up with the same API responses as you were, or Claude has hallucinated some extra field names.

I think the simplest steps forward would be:

1. Would you mind giving it a rebase off of the main branch, as there are some fixes here which have already been merged?
2. Are you able to add a new [vcr tests](https://docs.ropensci.org/vcr/) recording the batch API call?  An example is the Anthropic one [here](https://github.com/tidyverse/ellmer/blob/main/tests/testthat/_vcr/anthropic-batch.yml), and you can see how it's used in a test [here](https://github.com/tidyverse/ellmer/blob/ce0e342e12e1ba347abe6d0a483e498b2bfcd394/tests/testthat/test-provider-claude.R#L1-L8).  Then we can use this to test the code against.

Give me a shout if you want a hand with any of these, or alternatively, I'm happy to jump in to finish this off if you don't have the time any more, as I appreciate the original PR was submitted a while ago - just let me know what's best :)

R/batch-chat.R
xmarquez: Done. After rebasing, this has been removed.

tests/testthat/test-batch-chat.R
xmarquez: Same, rebased and now should be gone.

R/batch-chat.R
xmarquez: Same with this.

R/batch-chat.R
xmarquez: All done too.

R/provider-google.R
xmarquez: I checked in 110 saved Gemini batch state files across three of my projects (~76,000 responses total, mix of gemini-flash-latest and gemini-3.1-pro-preview, thinking enabled, varied schemas) and it looks like we always find it only in metadata. So we don't need to check for both of these, I think; this was probably defensive coding by Claude from an earlier iteration.

R/provider-google.R
xmarquez: Done - have simplified.

R/provider-google.R
xmarquez: Have checked and now collapsed to just x$key regex + line-number fallback, matching your suggestion. I checked my  110-file corpus mentioned above, which shows this nothing else but `key`; this was probably again defensive coding from an earlier iteration that had the wrong wrapper format. I also ran a live test (5 prompts, gemini-flash-latest, thinking on,  structured output) to confirm and it seems fine to simplify, thank you.

R/provider-google.R
xmarquez: Adopted your snippet, thank you.

R/provider-google.R
xmarquez: `has_batch_support` now returns `identical(provider@name, "Google/Gemini")`, so Vertex should return `FALSE`. Updated the corresponding test to expect this. I agree, the Vertex API seems completely different.

xmarquez: Also removed - checked my projects and yes, the promptFeedback-only "blocked content" shape appears in exactly 2 / 75,815 responses (both PROHIBITED_CONTENT), and in both cases it arrives wrapped in `x$response` so it should be ok to remove.

xmarquez:
Thanks for the review! I rebased on main and applied your suggested simplifications. Before changing the response-parsing helpers I had Claude and Codex scan 110 saved batch states from three of my projects (~76k responses) to confirm, then ran an end-to-end live test against gemini-flash-latest with thinking + structured output across 5 prompts from one of these projects to confirm it all seems to work.

One note: I haven't done the vcr tests yet. Still working on that.

R/provider-google.R
xmarquez: Deleted, you're right.

xmarquez:
Hi @thisisnic - I've tried to create the vcr tests recording the batch API call, but it seems to require me to modify `gemini_dowload_file()` for it to work. According to Claude, "vcr's request/response model captures the response body in-memory; `httr2::req_perform(req, path = path)` is the streamed-to-file form, and on cassette replay vcr returns the in-memory body without honoring the path arg, so the download file is never written and downstream `read_ndjson()` fails." Claude suggests a refactor that reads the response body into memory (`resp_body_raw()`) and writes it via `writeBin()` instead. 

My understanding is that `anthropic-batch.yml` is a single /v1/messages interaction (plain non-batch chat completion - it doesn't include the full batch flow; maybe it was intended as a vcr casette of the full batch flow but it was simplified at some point?), and the actual batch chat test in `test-provider-claude.R` uses a saved state fixture at `batch/state-capitals-anthropic.json` rather than vcr. We already have the Gemini equivalent: `batch/state-capitals-gemini.json` and the batch chat test works with that Gemini fixture at `tests/testthat/test-provider-google-batch.R`.

Did you want a full vcr cassette of the multi-step batch HTTP flow (upload, then submit, poll loop, download)? The latter seems doable (Claude created one, recording the full upload, submit, repeated polls, and final download - 44 interactions total, including 41 GETs.) but the download step uses `httr2::req_perform(path = ...)` which vcr's request/response model doesn't transparently capture, so "a vcr-based test would also need a small refactor of `gemini_download_file()` to read the body in-memory before writing". Let me know what you'd prefer me to do! I also checked via Codex and both models agree (Codex as usual did some extensive testing); a small refactor of `gemini_download_file()` seems to be necessary.

R/provider-google.R
thisisnic: I'm making a quick note here in case anyone is looking to implement structured output for the Vertex API later and is looking at this PR to get started.

I had Claude take a look at the [Python GenAI SDK library](https://github.com/googleapis/python-genai) which implements *both* the Gemini Developer API and the Vertex AI API, so I could be more sure that the differences prevent us just using the same code path here.  Claude's summary showed that Vertex uses `batchPredictionJobs` not `batchGenerateContent`, and the output goes to GCS/BigQuery, not a `responsesFile` so we'd need to add some code to implement that.  The Python GenAI SDK is pretty new, and looks like a good resource for that comparison.

thisisnic:
> My understanding is that `anthropic-batch.yml` is a single /v1/messages interaction (plain non-batch chat completion - it doesn't include the full batch flow; maybe it was intended as a vcr casette of the full batch flow but it was simplified at some point?), and the actual batch chat test in `test-provider-claude.R` uses a saved state fixture at `batch/state-capitals-anthropic.json` rather than vcr. We already have the Gemini equivalent: `batch/state-capitals-gemini.json` and the batch chat test works with that Gemini fixture at `tests/testthat/test-provider-google-batch.R`.

Yeah, we should be good with things as they are!

thisisnic (approved):
Thanks for sticking with this, @xmarquez! I made a small tweak to this, but should be good to go.

xmarquez:
Thank you @thisisnic! I've rebased on main and force-pushed. Please let me
know if you need anything else!

thisisnic:
Thanks @xmarquez!

PR #927, opened 2026-02-14
Add Groq batch processing support

### LINKED ISSUE #914: Feature Request: add Google Gemini support batch_chat()

xmarquez:
## Summary

Adds batch processing support for `chat_groq()` via the Groq Batch API (50% cost discount, 24h completion window). Closes #914 (Groq portion — Gemini batch was #926).

Implements six batch methods on `ProviderGroq` (`has_batch_support`, `batch_submit`, `batch_poll`, `batch_status`, `batch_retrieve`, `batch_result_turn`), reusing OpenAI's `openai_upload`, `openai_download_file`, and `openai_json_fallback` helpers (the bodies are identical).

## Implementation notes

- Mirrors the existing OpenAI batch JSONL format (`custom_id` / `method` / `url` / `body`) but targets Groq's `/v1/chat/completions` endpoint.
- `batch_status` maps Groq's terminal states (`completed`, `failed`, `expired`, `cancelled`) to `working = FALSE` — OpenAI's bare `!= "completed"` check would busy-poll forever on a failed/expired Groq batch.
- `batch_retrieve` guards `output_file_id` / `error_file_id` with `length() == 1 && nzchar()` because Groq returns `"error_file_id": {}` (empty JSON object) for successful batches.
- `batch_result_turn` uses Gemini's safer guard (`!is.null(result) && result$status_code == 200L && !is.null(result$body)`) so failed/expired rows from the error file don't crash.

## Tests

Two fixture-based tests in `tests/testthat/test-provider-groq.R`:
- `batch chat works` — text path, `llama-3.1-8b-instant`
- `batch_chat_structured works` — structured path, `openai/gpt-oss-20b`

Fixtures (`tests/testthat/batch/state-capitals-groq.json`, `state-capitals-groq-structured.json`) were recorded against live Groq.

## Test plan

- [x] `devtools::test()` — 0 failures across the batch suite
- [x] `devtools::check()` — 0 errors, 0 warnings
- [x] Verified live end-to-end: submit → poll → download → parse against Groq's API
- [x] Test scope follows the trimmed pattern from #996


thisisnic (changes requested):
Thanks for submitting this PR @xmarquez.  Is it possible to separate this out into 2 PRs; one for batch processing and one for structured output? Splitting them out would make it easier to review.

For the batch processing PR, would you mind refactoring the tests to make them more in line with the existing providers tests - the Claude and OpenAI tests are probably good examples, but there might be others.

xmarquez:
Thanks @thisisnic . I've force-pushed a single commit that drops the structured output parts entirely (seems to be redundant given PR #934, which appears to be merged already, according to Claude). I tried to make the tests match the Gemini patterns (not using a lot of mocking), and added two fixtures: a plain text one (one using llama-3.1-8b-instant) and a structured one (using openai/gpt-oss-20b, reasoning, structured outputs). Both fixtures come from live testing on Groq. Let me know what you think!

tests/testthat/test-provider-groq-batch.R
```diff
@@ -0,0 +1,153 @@
+# Batch support -----------------------------------------------------------
+
+test_that("ProviderGroq has batch support", {
+  chat <- chat_groq(credentials = function() list(Authorization = "Bearer x"))
+  expect_true(has_batch_support(chat$get_provider()))
+})
```
thisisnic: Some of these granular tests become redundant as `has_batch_support()` ends up being tested implicitly by other tests anyway, so we can remove them.

thisisnic (changes requested):
Thanks for the update @xmarquez! 

There quite a bit of duplicated code here; i.e. `groq_upload_file()` is identical to `openai_upload()` - can we reduce the duplication?

There are also likely still too many tests; in #996 I removed some of the earlier ones from the Google batch PR, just keeping the basic one and then tests of unusual behaviours we've needed to implement workarounds for.  I'm going to wait to get more input on that one about the level of testing we want, but once that's reviewed and merged, it can help guide us here.

xmarquez:
Thanks @thisisnic - I've deleted `groq_upload_file`, `groq_download_file`, and `groq_json_fallback` and replaced them with the corresponding `openai_*` helpers. I've also put all the batch tests into `test-provider-groq.R` - or rather, put the two fixture tests there and dropped all the other tests. (I deleted the live API test, the has_batch_support test, and the three batch_status tests - let me know if you think any of these should stay). I also did a quick live test on my end to make sure everything works (it still does), but I didn't include it.

xmarquez:
Sorry, one other question: do we need an additional test that checks multiple terminal states, like this:

```
test_that("batch_status recognizes terminal failure states", {
  provider <- chat_groq(
    credentials = function() list(Authorization = "Bearer x")
  )$get_provider()

  for (state in c("completed", "failed", "expired", "cancelled")) {
    batch <- list(
      status = state,
      request_counts = list(total = 2L, completed = 2L, failed = 0L)
    )
    expect_false(batch_status(provider, batch)$working)
  }
})
```

Claude and Codex think we might need it, since `BatchJob$wait()` loops while `batch_status(...)$working` is `TRUE`. So if Groq returns `failed`, `expired`, or `cancelled` and `batch_status()` only checks `status != "completed"`, `batch_chat(wait = TRUE)` can wait forever. Is this your read? I have not, however, adopted this suggestion yet.

thisisnic:
> Sorry, one other question: do we need an additional test that checks multiple terminal states, like this:
> 
> ```
> test_that("batch_status recognizes terminal failure states", {
>   provider <- chat_groq(
>     credentials = function() list(Authorization = "Bearer x")
>   )$get_provider()
> 
>   for (state in c("completed", "failed", "expired", "cancelled")) {
>     batch <- list(
>       status = state,
>       request_counts = list(total = 2L, completed = 2L, failed = 0L)
>     )
>     expect_false(batch_status(provider, batch)$working)
>   }
> })
> ```
> 
> Claude and Codex think we might need it, since `BatchJob$wait()` loops while `batch_status(...)$working` is `TRUE`. So if Groq returns `failed`, `expired`, or `cancelled` and `batch_status()` only checks `status != "completed"`, `batch_chat(wait = TRUE)` can wait forever. Is this your read? I have not, however, adopted this suggestion yet.

What are your thoughts on the reasons for/against doing this?

xmarquez:
> What are your thoughts on the reasons for/against doing this?

After thinking about it, I don't think it's worth adding - the implementation in `provider-groq.R` already is correct. I guess the only reason to keep it might be if anyone were to simplify the code there by trying to use the `provider-openai.R` implementation (which uses `working = batch$status != "completed"`) the test would catch this possibility.

thisisnic:
Yeah, I agree, could be too granular.  Have pulled this locally and tested it and it appears to be working - I'll make some time next week to finish reviewing.

thisisnic:
Still thinking about a few things around design - there's a lot of duplication between the Groq and the OpenAI (*not* OpenAI-compatible) batch processing, with the Groq code being more defensive, and I'm wondering if it's going to make more sense to just combine them and make a judgment call on how defensive we want the code to be versus overdoing it.  I think for the moment, it might be simpler to merge this with any small tweaks and then come back to possible refactoring at a later point.

thisisnic (approved):
Approved, pending green CI!

xmarquez:
Thank you @thisisnic, much appreciated. Let me know if I can help.

PR #934, opened 2026-02-21
Enable structured chat for Groq models

### LINKED ISSUE #930: Structured chat does not work with Groq models
Brand new to `ellmer`. This code seemed to work with other model providers, but not Groq:
```r
chat = chat_groq(model="openai/gpt-oss-20b")
chat$chat_structured("Is Boston the capital of MA?", type=type_boolean())
#> Error in `req_perform()`:
#> ! HTTP 400 Bad Request.
#> ℹ invalid JSON schema for response_format: 'structured_data':
#>   `additionalProperties:false` must be set on every object
#> Run `rlang::last_trace()` to see where the error occurred.
```

Happy to submit a PR if this is e.g. a recent change in their API.

CoryMcCartan:
This PR fixes #930.

I added a test for this use case, which requires using a different model than the default.  The existing test doesn't require a different default, so I wasn't sure if I should add a `chat_groq_test()` function like some other providers have. I also notice the tool-calling test doesn't seem to reliably pass with the default model.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* `chat_groq()` now supports structured chat (@CoryMcCartan, #930).
 * ellmer will now distinguish text content from thinking content while streaming, allowing downstream packages like shinychat to provide specific UI for thinking content (@simonpcouch, #909).
 * `chat_github()` now uses `chat_openai_compatible()` for improved compatibility, and `models_github()` now supports custom `base_url` configuration (@D-M4rk, #877).
 * `chat_ollama()` now contains a slot for `top_k` within the `params` argument (@frankiethull).


R/provider-groq.R
```diff
@@ -99,7 +93,8 @@ method(as_json, list(ProviderGroq, TypeObject)) <- function(provider, x, ...) {
     type = "object",
     description = x@description,
     properties = as_json(provider, x@properties, ...),
-    required = as.list(names2(x@properties)[required])
+    required = as.list(names2(x@properties)[required]),
```
hadley: Would you mind checking to see if it's now possible to just delete this method? i.e. maybe grok is now sufficiently OpenAI compatible that we don't need any customisation here.

CoryMcCartan: Yes, it looks like both this method and the `as_json` method for `Turn` objects are no longer needed.  The tool calling and structured data tests pass when using the default OpenAICompatible provider

hadley: Yay, deleting code is my favourite 😄

hadley:
One last thing and we should be good to merge:  Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

CoryMcCartan:
Thanks for your patience!

hadley:
Thank you!

PR #935, opened 2026-02-28
Drop empty ContentText in OpenAI-compatible serialization

### LINKED ISSUE #932: chat_databricks(): HTTP 400 when conversation history contains empty ContentText
## Problem

When the conversation history contains a `ContentText("")` (empty text), `chat_databricks()` fails with HTTP 400 because the Databricks Claude proxy rejects empty text content blocks.

This can happen organically during tool calling when a model responds with tool calls and an empty `content` field (`content: ""` instead of `content: null`). When ellmer replays the conversation history on the next round-trip, the serialized empty text block `{"type": "text", "text": ""}` is rejected.

## Reprex

```r
library(ellmer)

chat <- chat_databricks(model = "databricks-claude-sonnet-4-5")

# Simulate what happens when the model returns content = "" during tool calling
chat$set_turns(list(
  UserTurn(list(ContentText("Hello"))),
  AssistantTurn(list(ContentText("")))
))

# Fails because the empty text content block is serialized as
# {"type": "text", "text": ""} which Databricks rejects
chat$chat("How are you?", echo = "none")
#> Error in `req_perform()`:
#> ! HTTP 400 Bad Request.
#> ℹ {"message":"messages: text content blocks must be non-empty"}
```

## Root cause

This is a Databricks-specific double problem:

1. **Databricks returns `content: ""` in responses** for tool-only turns (instead of `content: null` which OpenAI returns). This causes `value_turn()` for `ProviderOpenAICompatible` (`R/provider-openai-compatible.R:290`) to create `ContentText("")`:

```r
content <- lapply(message$content, as_content)
# lapply("", as_content) → ContentText("") since as_content() doesn't validate non-empty strings
```

2. **Databricks then rejects `content: ""` when replayed back** → HTTP 400.

Other OpenAI-compatible providers (OpenAI, Azure OpenAI) return `content: null` for tool-only turns, so `lapply(NULL, as_content)` → `list()` and no empty `ContentText` is ever created. Databricks creates the problem and then rejects it.

## Possible fixes

1. **Filter in `value_turn()`**: Drop empty `ContentText` when parsing model responses
2. **Filter in `as_json()`**: Skip empty `ContentText` during serialization (similar to the existing empty assistant turn check in `provider-claude.R:418-422`)
3. **Validate in `ContentText`**: Reject empty strings at construction time (breaking change)

JamesHWade:
## Summary

- Fixes `chat_databricks()` HTTP 400 errors when conversation history contains empty `ContentText("")` objects, which occur during tool calling when Databricks returns `content: ""` instead of `content: null`
- Filters out empty `ContentText` during `as_json()` serialization for `ProviderOpenAICompatible`, following the existing pattern from the Claude provider
- Drops entirely empty assistant turns (after filtering) to avoid API errors

Closes #932

## Test plan

- [x] Added unit tests verifying empty `ContentText` is dropped during serialization
- [x] Added test for assistant turns with only empty `ContentText` (dropped entirely)
- [x] Added test for mixed content (empty text stripped, real content preserved)
- [x] Added test for empty text with tool requests (text stripped, tools preserved)
- [x] Existing test suite passes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # ellmer (development version)
 
 * `chat_databricks()` (and other `chat_openai_compatible()` providers) no longer fail with HTTP 400 when the conversation history contains empty `ContentText("")` objects, which can occur during tool calling (@JamesHWade, #932).
+* `chat_groq()` now supports structured chat (@CoryMcCartan, #930).
 * ellmer will now distinguish text content from thinking content while streaming, allowing downstream packages like shinychat to provide specific UI for thinking content (@simonpcouch, #909).
 * `chat_github()` now uses `chat_openai_compatible()` for improved compatibility, and `models_github()` now supports custom `base_url` configuration (@D-M4rk, #877).
 * `chat_ollama()` now contains a slot for `top_k` within the `params` argument (@frankiethull).
--- a/R/provider-groq.R
+++ b/R/provider-groq.R
@@ -8,10 +8,6 @@ NULL
 #'
 #' Built on top of [chat_openai_compatible()].
 #'
-#' ## Known limitations
-#'
-#' groq does not currently support structured data extraction.
-#'
 #' @export
 #' @family chatbots
 #' @param api_key `r lifecycle::badge("deprecated")` Use `credentials` instead.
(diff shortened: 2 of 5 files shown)


R/provider-openai-compatible.R
```diff
@@ -343,10 +343,21 @@ method(as_json, list(ProviderOpenAICompatible, Turn)) <- function(
...
+    # Drop empty ContentText to avoid API errors from providers that reject
+    # empty text content blocks (e.g. Databricks returns content: "" for
+    # tool-only turns, then rejects it on replay)
+    is_empty_text <- map_lgl(x@contents, function(c) {
+      S7_inherits(c, ContentText) && !nzchar(c@text)
+    })
+    contents <- x@contents[!is_empty_text]
+    if (length(contents) == 0) {
+      # Drop empty assistant turns to avoid an API error
+      return(list())
```
hadley: I think this should probably be a `NULL`?

tests/testthat/test-provider-openai-compatible.R
```diff
@@ -92,6 +92,38 @@ test_that("structured data work with and without wrapper", {
...
+  result <- as_json(stub, turn)
+  expect_equal(
+    result,
+    list(list(
+      role = "assistant",
+      content = list(list(type = "text", text = "Hello"))
+    ))
+  )
+
+  # Empty ContentText is stripped but tool requests are preserved
```
hadley: This should be a separate test I think

tests/testthat/test-provider-openai-compatible.R
```diff
@@ -92,6 +92,38 @@ test_that("structured data work with and without wrapper", {
...
+  ))
+  result <- as_json(stub, turn)
+  expect_equal(
+    result,
+    list(list(
+      role = "assistant",
+      content = list(list(type = "text", text = "Hello"))
+    ))
+  )
+
```
hadley, line 150: I think you're missing a test for the `(length(contents) == 0)` case? i.e. something like `list(ContentText(""), ContentText(""))`. I'd also like to see a live test for that because my recollection from doing this work on Claude is that will cause a problem.

JamesHWade:
Thanks for the review! Addressed all three points:

1. Changed `return(list())` to `return(NULL)` for consistency
2. Split the tool request test into its own `test_that()`
3. Added test for the all-empty case (`list(ContentText(""), ContentText(""))`) and a live Databricks test

Also fixed the root cause: `value_turn()` now treats `content: ""` as `content: null`, so `ContentText("")` is never created from provider responses in the first place. The stripping in `as_json()` remains as a safety net for manually constructed turns.

hadley:
Thank you!

PR #939, opened 2026-03-14
Add chat_bedrock_anthropic() with Anthropic prompt caching support

1beb:
Uses Bedrock's InvokeModel API with Anthropic Messages format instead of the Converse API, enabling cache_control on system prompts and message content. Inherits from ProviderAnthropic for full feature parity including streaming, tool use, and token tracking.

I've tested this against a production system at work. It does it's job of caching a rather large system prompt. There are a couple of magic strings in here that I hate: 

```  
# Add Bedrock-required anthropic_version
body$anthropic_version <- "bedrock-2023-05-31"
```

Who knows when that will change? It's specified in the example here: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

Your version of anthropic provider has a custom is_transient function that handles 429, 503, 529. I just used ellmer_req_robustify() here for consistency with the other methods. That means it would fail instead of retrying on 503, 529. Bedrock throttles with 429 so I don't think it matters. Easy fix though.


hadley:
@gadenbuie do you have a little time to review/test this?

gadenbuie:
Thanks @1beb! We've been wanting to do this for a while and I appreciate you taking a pass at it. I did a little bit of research and I think that we can achieve similar results by inserting `cachePoint` objects in the JSON payload with the `Converse` and `ConverseStream` APIs. I wrote up my research and plan in #954. 

Since you looked into this and used the `InvokeModel` API, please let me know if you see any issues with that approach or if you've already encountered scenarios where the `InvokeModel` API works better than the Converse APIs.

1beb:
Actually, I think your approach might be better. I chose InvokeModel because my use case was not a multi-turn convo but instead a single completion that's largely stateless but repetitive with a big cache hit. Picture an extra-large system prompt with a very small user prompt 1000s of times. 

That's probably not the right approach for ellmer generally (vs. my person use-case). 

I'm trying to read through the docs and the tradeoffs are pretty nominal: 

- payload size can be a bit bigger with invokemodel
- Some extra json to parse messages, role, content in Converse 
- A little unclear how to submit model params vs. Invoke. I think it's [additionalModelRequestFields](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#bedrock-runtime_Converse-request-additionalModelRequestFields) 

I have to do another set of Opus runs soon using Bedrock, I'll see if I can switch to Converse. 

Let's close this PR for now, I'll reference it when I swap to Converse.

PR #943, opened 2026-03-16
Fix gargle credentials skipped in interactive mode

### LINKED ISSUE #922: Vertex AI authentication via application credentials not working in interactive mode
Hi, 

I'm trying to use Gemini through Google Vertex AI with service account credentials. According to the [documentation](https://ellmer.tidyverse.org/reference/chat_google_gemini.html#authentication), ellmer should use Google's default application credentials if available, but in interactive mode it always triggers the browser-based OAuth flow instead.

```r test.R

library(ellmer)
library(gargle)

Sys.setenv(
  "GOOGLE_APPLICATION_CREDENTIALS" = "my-demo-creds.json"
)

# Create chat client
chat_client <- ellmer::chat_google_vertex(
  location = "europe-west3",
  project_id = "my-demo-project",
  model = "gemini-2.5-flash",
  echo = "none"
)

# Test chat
response <- chat_client$chat("Say 'Hello from non-interactive R!'")
cat(response, "\n")

```

When run in non-interactive mode in the command line (e.g., via Rscript test.R), this script works fine and uses the application credentials. However, in interactive mode (RStudio, Shiny), it triggers the OAuth browser flow instead.

Is it possible to use application credentials in interactive mode? This would be especially useful when using ellmer in a Shiny app via shinychat.

stefanlinner:
Closes #922.

In `default_google_credentials()`, the `is_interactive()` check on line 733 unconditionally returned the OAuth browser flow, even when `gargle::token_fetch()` had already obtained a valid token from application default credentials (e.g. via `GOOGLE_APPLICATION_CREDENTIALS`). This meant that in interactive sessions (RStudio, Shiny), service account credentials were silently ignored.

The fix adds an `is.null(token)` guard so the OAuth fallback only activates when no token was obtained via gargle. This matches the [documented authentication order](https://ellmer.tidyverse.org/reference/chat_google_gemini.html#authentication) and is consistent with how the Azure and Databricks providers handle their interactive OAuth fallbacks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* `default_google_credentials()` no longer skips application default credentials (e.g. `GOOGLE_APPLICATION_CREDENTIALS`) in interactive sessions, instead falling through to the OAuth browser flow only when no gargle token is available (@stefanlinner, #922).
 * `chat_databricks()` (and other `chat_openai_compatible()` providers) no longer fail with HTTP 400 when the conversation history contains empty `ContentText("")` objects, which can occur during tool calling (@JamesHWade, #932).
 * `chat_groq()` now supports structured chat (@CoryMcCartan, #930).
 * ellmer will now distinguish text content from thinking content while streaming, allowing downstream packages like shinychat to provide specific UI for thinking content (@simonpcouch, #909).


hadley:
Thank you!

Issue #585, opened 2025-06-23
chat_azure_ai_foundry function for models deployed on Azure AI Foundry


pietervreeburg:
My (local government) organisation uses Azure AI services for hosting LLM's from different providers. I've succeeded in configuring ellmer to use gpt-4.1 on an Azure OpenAI resource, but I've not yet managed to get ellmer working with a model from Mistral deployed on an Azure AI Foundry resource. I've tried tweaking the defaults of the `chat_azure_openai` function to suit my needs, but so far I haven't produced anything working.

Are you planning to add a `chat_azure_ai_foundry` function (or something like that) in the near future to ellmer? That would be very helpful to government workers like me who are more or less chained to the Microsoft ecosystem.

Any guidance on how to configure other functions in the ellmer package for use with Azure AI Foundry would be super helpful too.

pietervreeburg:
Using the code examples provided on the Azure AI Foundry portal and httr2 (nice package). I put together some code to test the connection and check whether I can get some results back from a non-OpenAI model hosted on my personal AI Foundry instance. This actually worked right out of the gate. I have posted the code below and will check, when I have some spare time, whether I can modify the existing chat_azure_openai function to connect to my Azure AI Foundry instance.

```
library(httr2)

req <- request("https://<YOUR-RESOURCENAME-HERE>.services.ai.azure.com/models/chat/completions?api-version=2024-05-01-preview")
req

req <- req %>%
  req_headers(
    `Content-Type` = "application/json",
    Authorization = api_key
  ) %>%
  req_body_raw(
    '{
      "messages": [
        {
          "role": "user",
          "content": "I am going to Amsterdam, what should I see?"
        }
      ],
      "max_tokens": 2048,
      "temperature": 0.8,
      "top_p": 0.1,
      "model": "Mistral-Large-2411"
    }'
  )

req

results <- req %>%
  req_perform() %>%
  resp_body_json()

results %>% 
  str()

```

hadley:
We don't have any Azure AI foundry users locally, so if you figure out how to modify `chat_azure_openai()` we'd love to know 😄

PR #960, opened 2026-04-10
Add `chat_azure_anthropic()` for Azure Foundry Anthropic Models

dareneiri:
# Summary

A new function `chat_azure_anthropic()` allows users to access Anthropic models made available through Azure Foundry, which has a different access point from the existing Azure OpenAI implementation in `chat_azure_openai()`

Azure AI Foundry exposes Anthropic models via a different endpoint format than `chat_azure_openai()`, which targets `*.openai.azure.com` and uses the OpenAI chat completions format. This new function targets the Foundry endpoint and uses the standard Anthropic Messages API format (`x-api-key`, `anthropic-version: 2023-06-01`), meaning it inherits all existing `ProviderAnthropic` behaviour — streaming, tool calling, structured data extraction, prompt caching, batch chat, and image/PDF support — with no additional implementation required.

# Changes

* **New function:** `chat_azure_anthropic()` which connects to Anthropic models hosted at`*.services.ai.azure.com/anthropic` 
* Adds new line to news.md on new function

# Implementation details

* `ProviderAzureAnthropic` inherits from `ProviderAnthropic` with no additional properties — the Foundry Anthropic endpoint uses the same API format as api.anthropic.com 
* `model` is required with no default, since the value is a user-specific Azure deployment name rather than a canonical model identifier, similar to Azure Open AI
* Trailing slashes on endpoint are normalised to avoid double-slash URLs
* `default_azure_anthropic_credentials()` mirrors `default_azure_credentials()` with the same priority order

# Testing

* ✅  New test added: `tests/testthat/test-provider-azure-anthropic.R`
```
devtools::test(filter = "provider-azure-anthropic")
# [ FAIL 0 | WARN 0 | SKIP 1 | PASS 15 ]
# SKIP: connectcreds not installed (expected)
```

```
> chat <- chat_azure_anthropic(model = "claude-sonnet-4-6")
> chat$chat("What is the capital of France?")
The capital of France is **Paris**.
```


dareneiri:
I previously wrote that this addresses #585 but that specifically mentions the ability to call Mistral models through Azure Foundry. This new function only concerns Anthropic models. Mistral models will call `*.services.ai.azure.com/openai/v1/`.

PR #967, opened 2026-05-01
Automate weekly price data refresh via GitHub Action

gadenbuie:
Pairs with #968 which adds runtime price cache refresh in the package.

## Summary

- Adds a weekly GitHub Action (`.github/workflows/update-prices.yaml`) that fetches litellm pricing data, transforms it, validates against a JSON schema, rebuilds `sysdata.rda`, and commits the result
- `data-raw/prices.R` uses tidyverse (dplyr/tidyr/stringr) for transformation with `jsonvalidate` for schema validation (ajv engine)
- Adds `data-raw/prices.csv` for human-readable diffs on pricing changes
- Adds `Config/Needs/prices` to DESCRIPTION so CI only installs what the script needs
- Adds a JSON schema (`data-raw/prices.schema.json`) for validation
- Fixes unrelated test failures in `test-provider-claude.R` and `test-provider-google.R`

## Test plan

- [x] Verify `Rscript --quiet --vanilla data-raw/prices.R` produces valid `prices.json` and `prices.csv`
- [x] Verify GHA workflow runs correctly and skips on forks
- [x] Confirm existing tests still pass with updated `sysdata.rda`


hadley:
Could we split this into two PRs? I think the "Price data pipeline" will be straightforward to review and merge, but I'll want to think more about the run-time cache.

hadley:
Can you make the case for why we should do this in jq + bash instead of the existing R script? While I'm sure that R code is not the most efficient, I don't think taking even an extra minute or two is going to make a meaningful if we're running this on CI, and whereas before I could easily understand what the code was doing at a glance, now I'd need to do some work with claude.

gadenbuie:
> Can you make the case for why we should do this in jq + bash instead of the existing R script? While I'm sure that R code is not the most efficient, I don't think taking even an extra minute or two is going to make a meaningful if we're running this on CI, and whereas before I could easily understand what the code was doing at a glance, now I'd need to do some work with claude.

The biggest reason I did this was to get easy validation via JSON schema so that we'd feel better about having this file update automatically. Secondarily, my goal was to make the CI run quickly; installing R is a heavy dependency compared to the fact that a bash and Python implementation of the same process runs in roughly 10 seconds. We still install R if the data has changed, but the goal was to feel better about setting up a workflow that runs routinely. Finally, I felt the process was stable enough and the resulting jq code was readable enough. 

Still, I understand where you're coming from in terms of overall accessibility and I'm fine with going back to the original version running on a schedule. I can get this quickly turned around tomorrow morning if you want.

hadley:
Yeah, I don't think the overhead is significant enough to worry about, especially given that it's usually going to be run attended in CI. So I would prefer to stick with R. Thanks for being flexible with this!

gadenbuie:
@hadley I restored the R-based processing, keeping the JSON schema validation though jsonvalidate. This should be ready for your review now!

Oh, I also have the workflow write out `data-raw/prices.csv` just to give us diffable changes on what we ship with the package. There's some churn there to sort the rows deterministically, but in the future we should get clean diffs from the process.

.github/workflows/update-prices.yaml
```diff
@@ -0,0 +1,58 @@
+on:
+  push:
+    paths:
+      - .github/workflows/update-prices.yaml
+  schedule:
+    - cron: "0 6 * * 1"
```
hadley: IMO it's good practice to include a comment here explaining when this will be run

.github/workflows/update-prices.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: release
+          use-public-rspm: true
+
+      - name: Install system dependencies
+        run: sudo apt-get install -y libnode-dev
+
+      - name: Install dependencies
```
hadley, line 29: https://github.com/r-lib/actions/tree/v2-branch/setup-r-dependencies didn't work here?

data-raw/prices.R
```diff
@@ -31,34 +34,58 @@ all_prices <- df |>
...
+  engine = "ajv",
+  error = TRUE
+)
+cli::cli_progress_done()
+
+# --- write outputs -----------------------------------------------------------
 
-# Output JSON for ingestion into `chatlas`
 jsonlite::write_json(prices, "data-raw/prices.json", pretty = TRUE)
+write.csv(prices, "data-raw/prices.csv", row.names = FALSE)
```
hadley: You think this is more helpful than the json file? If so, do we still need to write the json? I think I was thinking of that as the readable diff.

hadley (approved):

.github/workflows/update-prices.yaml
gadenbuie:
```suggestion
    - cron: "0 6 * * 1" # Monday at 6am UTC
```

.github/workflows/update-prices.yaml
gadenbuie: No, I couldn't figure out the syntax and ran out of patience after a few tries. It would work fine if we also installed all of ellmer's dependencies, though. I'm probably trying too hard to reduce CI overhead, but we only need a couple of packages.

The other option is to use `setup-r-dependencies` but _not_ use `Config/Needs/prices`.

data-raw/prices.R
gadenbuie: yeah good point, the JSON is fine. I'll clean this up

PR #968, opened 2026-05-04
Add `models_update_prices()` to refresh cached pricing data from GitHub

gadenbuie:
Builds on #967, which added the GHA/data-raw infrastructure.

## Summary

Adds a user-controlled caching layer (`R/prices.R`) that keeps model pricing data fresh without requiring a package update. Users call `models_update_prices()` to download and cache the latest `prices.json` from GitHub; `prices_get()` then reads the cached data in place of the bundled snapshot.

- Cached prices are stored as RDS in `tools::R_user_dir("ellmer", "cache")`
- Downloads go through httr2 with `req_cache()`, so an unchanged upstream file costs at most a conditional request
- `prices.json` is now a versioned envelope (`schema_version`, `min_ellmer_version`, `data`); incompatible cached files are detected and discarded rather than used
- Informs the user with a clickable link when pricing data is actually updated

## Key design decisions

- **`models_update_prices()` is user-initiated**: There is no automatic background refresh. Users call `models_update_prices()` when they want to pull the latest pricing data; ellmer's bundled snapshot covers the common case between updates.
- **Conditional requests via `httr2::req_cache()`**: httr2 handles the ETag/304 round trip, so a call that finds nothing new is cheap. Whether the data actually changed is decided by comparing the parsed data against what's already cached, not by the status code.
- **Cached data replaces bundled data wholesale**: `prices.json` is the complete dataset, not a delta, so a compatible cache is a full replacement. `prices_get()` returns cached-or-bundled with no row-level merge.
- **Releases are a ratchet**: the cache records the ellmer version that wrote it. Every release regenerates the bundled snapshot, so a cache written by an older ellmer has been superseded and is silently ignored in favor of the bundled data. Development versions don't tick with each snapshot, so they keep using whatever they cached — dev users get the friction, releases stay correct. Upgrading and re-running `models_update_prices()` re-stamps the cache even when the pricing data itself hasn't changed, so a cache never stays dead waiting for upstream to move.
- **Schema versioning guards cache compatibility**: `prices.json` carries a `schema_version` integer. `prices_get()` rejects cached data whose version differs from the bundled data and falls back gracefully with an informative message. `prices_cache_download()` rejects remote data whose version differs from the bundled data with a clear error.
- **Validation is asymmetric on purpose**: `data-raw/prices.R` is the producer and validates strictly (JSON schema plus row and provider counts) because it defines the shape. The checks in `R/prices.R` are the consumer side and are deliberately looser — enough for an installed ellmer to decide whether a newer `prices.json` is safe to read, assuming strict validation already happened upstream.
- **Informative errors on download failure**: Network errors, HTTP errors, malformed JSON, and schema mismatches all surface as actionable `cli::cli_abort()` messages rather than silent failures.
- **Eager cache population after update**: `models_update_prices()` calls `prices_get()` immediately after clearing `the$prices` so the new data is loaded into the session cache before returning.

## Note for reviewers

`prices.json` on `main` is still the pre-envelope bare array, so `models_update_prices()` will error against the live URL until this merges and the workflow regenerates the file. The happy path was verified against this branch's `prices.json`.

## Tests included

- [x] `prices_get()`: uses cached data in place of bundled data; uses bundled prices when no cache exists; errors when cached data is missing bundled columns
- [x] `prices_get()` schema version: informs and falls back to bundled when cached version is older; warns and falls back when cached version is newer
- [x] `prices_get()` release ratchet: ignores a cache stamped with an older ellmer version, ignores a cache with no stamp, honors a cache stamped newer (i.e. after a downgrade); `prices_cache_download()` records the stamp
- [x] `models_update_prices()`: informs and returns TRUE when download succeeds (cache pre-populated); informs and returns FALSE when already up to date
- [x] `prices_cache_download()` at the httr2 boundary, via `local_mocked_responses()`: 200 writes cache with `schema_version`; unchanged data returns FALSE; HTTP error aborts; network failure aborts; malformed JSON aborts; missing envelope fields abort; remote schema newer than bundled aborts with upgrade message; remote schema older than bundled aborts; missing required columns abort; non-numeric input/output abort
- [x] End-to-end against the real URL: `req_cache()` writes its cache and the envelope check rejects `main`'s current bare-array `prices.json`


R/prices.R
```diff
@@ -0,0 +1,191 @@
...
+  if (is.null(the$prices)) {
+    bundled <- prices_data
+    cached <- prices_cache_read()
+
+    if (is.null(cached)) {
+      the$prices <- bundled
+      return(invisible(the$prices))
+    }
+
+    if (is.null(cached$variant)) {
```
hadley: Why would this be true?

R/prices.R
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    key_cols <- c("provider", "model", "variant")
+    bundled_in_cache <- !is.na(vctrs::vec_match(
+      bundled[key_cols],
+      cached[key_cols]
+    ))
+    bundled_only <- bundled[!bundled_in_cache, ]
+    # Align columns before rbind in case cached RDS is from an older schema
+    cached <- cached[, intersect(names(bundled), names(cached)), drop = FALSE]
```
hadley: We'd also need to ensure that the column types never change. 

I assume that's why you're doing the schema validation in the other PR. But I feel like the ideal place for the schema validation is here.

I think a better solution would be to explicitly version the prices data and then only combine if the version is as expected.

R/prices.R
```diff
@@ -0,0 +1,191 @@
...
+
+  resp <- tryCatch(
+    curl::curl_fetch_memory(url, handle = handle),
+    error = function(e) NULL
+  )
+  if (is.null(resp)) {
+    return(FALSE)
+  }
+  if (resp$status_code == 304L) {
+    return("not_modified")
```
hadley: This makes the return type a bit surprising

hadley (commented):
I like the idea of this PR and feels like you've protected against all the expected failure modes. What I worry is the unknown unknowns: there are lots of things that might go wrong when downloading data. And we're particularly vulnerable to them since we need to do the work in the main thread. I worry that curl is on the critical path here, and in the past I have found myself surprised by exactly how the timeout parameters work. I also worry about how this automated downloading of data might be perceived by strict enterprise IT departments, as it's a bit surprising for a package to download new data off the internet.

So overall, I'd love for us to be able to do this, but it just feels too high risk to me, given that the gain is relative small (and as far as I can tell, no one has actually complained about this). We could also reduce the need for it by committing to regular ellmer patch releases.

gadenbuie:
> So overall, I'd love for us to be able to do this, but it just feels too high risk to me, given that the gain is relative small (and as far as I can tell, no one has actually complained about this). We could also reduce the need for it by committing to regular ellmer patch releases.

What do you think about having a separate function, like `ellmer::model_update_prices()`? That would make it entirely an opt-in choice, and we could remove a lot of the extra machinery.

Personally, I've found this to be annoying enough when I'm using ellmer that I'm still pretty committed to trying to fix it, even if I'm the only one complaining.

R/prices.R
gadenbuie: Great point, this is a bit of a consequence of separating the PRs. I think it's a great idea to ship the schema with the package and then have the price updating process also validate before accepting the update

hadley:
Yeah, `ellmer::model_update_prices()` would be way more compelling.

R/prices.R
gadenbuie: I misread the original comment; versioning `prices.json` is a great idea and I implemented that in follow up

R/prices.R
gadenbuie: Great point, I went with just throwing from here rather than trying to coordinate with the caller.

R/prices.R
```diff
@@ -0,0 +1,229 @@
...
+    key_cols <- c("provider", "model", "variant")
+    stopifnot(
+      "cached pricing data is missing columns from bundled data" = all(
+        names(bundled) %in% names(cached)
+      )
+    )
+    bundled_in_cache <- !is.na(vctrs::vec_match(
+      bundled[key_cols],
+      cached[key_cols]
+    ))
```
gadenbuie:
```suggestion
    bundled_in_cache <- !is.na(
      vctrs::vec_match(bundled[key_cols], cached[key_cols])
    )
```

gadenbuie:
@hadley @thisisnic I've just merged in main; this is ready for a second look

R/prices.R
```diff
@@ -0,0 +1,228 @@
...
+# 3. Read: `prices()` merges cached + bundled rows, with cached winning
+#    on (provider, model, variant) conflicts. The result is memoized in
+#    `the$prices`; `models_update_prices()` clears it after a refresh.
+#
+# Both reads and writes are gated by an integer `schema_version`; see
+# `data-raw/prices.R` for the contract and when to bump it.
+
+prices <- function() {
+  if (is.null(the$prices)) {
+    bundled <- prices_data
```
hadley: Probably worth pulling this out into its own function

R/prices.R
```diff
@@ -0,0 +1,228 @@
...
+
+  handle <- curl::new_handle()
+
+  etag <- prices_cache_etag()
+  if (!is.null(etag)) {
+    curl::handle_setheaders(handle, `If-None-Match` = etag)
+  }
+
+  call <- rlang::caller_env()
+
```
hadley:
```suggestion
prices_cache_download <- function(call = caller_env()) {
  force(call)
  url <- "https://raw.githubusercontent.com/tidyverse/ellmer/refs/heads/main/data-raw/prices.json"

  handle <- curl::new_handle()

  etag <- prices_cache_etag()
  if (!is.null(etag)) {
    curl::handle_setheaders(handle, `If-None-Match` = etag)
  }

```

R/prices.R
```diff
@@ -0,0 +1,228 @@
...
+#' @return Invisibly returns `TRUE` if the cache was updated, or `FALSE` if
+#'   the cached data was already up to date. Throws an error if the download
+#'   fails or if the \pkg{curl} package is not installed.
+#' @export
+models_update_prices <- function() {
+  if (!requireNamespace("curl", quietly = TRUE)) {
+    cli::cli_abort(
+      "The {.pkg curl} package is required to update pricing data."
+    )
+  }
```
hadley:
```suggestion
```

Since httr2 depends on curl?

R/prices.R
```diff
@@ -0,0 +1,228 @@
...
+}
+
+curl_fetch_memory <- function(url, handle) {
+  curl::curl_fetch_memory(url, handle = handle)
+}
+
+prices_cache_download <- function() {
+  url <- "https://raw.githubusercontent.com/tidyverse/ellmer/refs/heads/main/data-raw/prices.json"
+
+  handle <- curl::new_handle()
```
hadley: Is it worth using httr2 here since it's already an ellmer dep? Then you could also take advantage of `httr2::req_cache()`

R/prices.R
```diff
@@ -0,0 +1,228 @@
...
+        call = call
+      )
+    }
+  )
+
+  if (!is.list(parsed) || !is.data.frame(parsed$data)) {
+    cli::cli_abort("Failed to parse pricing data from GitHub.", call = call)
+  }
+
+  remote_version <- as.integer(parsed$schema_version)
```
hadley, line 199: Can you share this error code with `prices.R`?

R/prices.R
gadenbuie: Done. `prices()` is now just the memoization, with `prices_merged()` doing the work and `prices_cache_compatible()` handling the schema check and its messaging.

R/prices.R
gadenbuie: Switched. It deleted more than I expected: the ETag storage, the `curl_fetch_memory()` shim I only had so the tests could mock curl, and all the manual status-code branching (httr2's own HTTP errors are better than what I was writing by hand). Tests now use `httr2::local_mocked_responses()`.

One behavioural change to be aware of: with `req_cache()` the 304 is handled inside httr2 and the cached body is replayed, so I can't observe "not modified" directly any more. `models_update_prices()` now decides whether anything changed by comparing the parsed data against what's already cached.

R/prices.R
gadenbuie: Want to make sure I'm reading this right before I move anything.

If you mean the required-columns / numeric-type checks: I'd rather keep them separate, because the asymmetry is deliberate. `data-raw/prices.R` is the producer and validates strictly (JSON schema + row and provider counts) because it *defines* the shape. The checks here are the consumer side and are intentionally much looser — they primarily exist so that an already-installed ellmer can tell whether a newer `prices.json` is still safe for it to read, with the assumption that strict validation was already handled upstream by the data-raw scripts.

If you mean the `schema_version` constant itself: that's currently defined in `data-raw/prices.R` and flows into the package as an attribute on `prices_data`, so a bump is a single edit in the place where the shape is actually decided. I could hoist it into `R/prices.R` instead, but that would move it away from where it changes and give us two things to keep in sync.

Or did you mean something else?

R/prices.R
```diff
@@ -0,0 +1,223 @@
...
+prices_merged <- function() {
+  bundled <- prices_data
+  cached <- prices_cache_read()
+
+  if (!prices_cache_compatible(cached, bundled)) {
+    return(bundled)
+  }
+
+  key_cols <- c("provider", "model", "variant")
+  stopifnot(
```
hadley: Weird to have a `stopifnot()` here. Maybe it can be moved into `prices_cache_compatible()`?

R/prices.R
```diff
@@ -0,0 +1,223 @@
...
+    "cached pricing data is missing columns from bundled data" = all(
+      names(bundled) %in% names(cached)
+    )
+  )
+  bundled_in_cache <- !is.na(
+    vctrs::vec_match(bundled[key_cols], cached[key_cols])
+  )
```
hadley:
```suggestion
  bundled_in_cache <- vec_in(bundled[key_cols], cached[key_cols])
```
?

R/prices.R
```diff
@@ -0,0 +1,223 @@
...
+  bundled_only <- bundled[!bundled_in_cache, ]
+  rbind(cached[names(bundled)], bundled_only)
```
hadley: It probably doesn't matter, but this changes the order of the rows. But it does make me wonder what we're trying to do here? Do we really need to _merge_ the bundled and cached? Can't we just take the new data?

R/tokens.R
```diff
@@ -114,18 +114,20 @@ token_usage <- function() {
...
 has_cost <- function(provider, model) {
   needle <- data.frame(provider = provider@name, model = model)
-  vctrs::vec_in(needle, prices[c("provider", "model")])
+  vctrs::vec_in(needle, prices()[c("provider", "model")])
 }
 
 get_token_cost <- function(provider, tokens, variant = "") {
   check_string(variant, .internal = TRUE)
 
+  p <- prices()
```
hadley: Maybe make it `prices_get()` or similar to avoid having to rename?

hadley (approved):
Feel free to merge after addressing the last few small things. Thanks!

R/prices.R
gadenbuie: Moved — it's now on the branch of `prices_cache_compatible()` where the cache is declared usable, which reads much better. It also matters more than it did, since we now hand the cached data back rather than merging with the bundled data.

R/prices.R
gadenbuie: Resolved automatically by using the cached data directly — see the thread below.

R/prices.R
gadenbuie: You're right, we don't. `data-raw/prices.R` writes the same frame to `prices.json` and to sysdata, so a compatible cache is a full replacement rather than a delta. `prices_get()` is now just:

```r
prices_get <- function() {
  if (is.null(the$prices)) {
    cached <- prices_cache_read()
    the$prices <- if (prices_cache_compatible(cached, prices)) cached else prices
  }
  the$prices
}
```

The one thing the merge bought us: a cache downloaded a while ago, paired with a freshly-updated ellmer, could be missing models the bundled snapshot has. That felt like the wrong trade to keep this machinery for — a stale cache is one `models_update_prices()` away, and the merge was arguably wrong in that case anyway (it let stale cached rows win over newer bundled prices).

R/tokens.R
gadenbuie: Done — the accessor is `prices_get()` and the internal data is back to `prices`, so `data-raw/prices.R` and the tests lose the rename churn.

R/prices.R
gadenbuie: Correcting myself on the last paragraph: I said the stale-cache case wasn't worth machinery, but it was cheap to handle more coarsely.

The cache now records the ellmer version that wrote it (ad08d77e). Since every release regenerates the bundled snapshot, a cache from an older ellmer is superseded by the data we ship, and `prices_get()` silently prefers bundled. Dev versions don't tick with each snapshot, so they keep whatever they cached — they're the ones who can afford the friction, and `models_update_prices()` is the way out.

So the ratchet gets us the freshness guarantee the merge was groping at, without the row-level matching or the reordering you flagged.

gadenbuie: Correction/update in f156e49c: the cache no longer records the version of the ellmer installation that downloaded it. Instead, `data-raw/prices.R` stamps each generated snapshot with the ellmer version from `DESCRIPTION`, and writes that producer version to both `prices.json` and bundled `prices` data. `prices_get()` compares those snapshot versions, so an older cached snapshot loses to the bundled snapshot. `models_update_prices()` skips a remote snapshot that is not newer than bundled data; when it is newer, it only writes if it differs from the existing cache. Thus a release uses its bundled snapshot without creating a redundant cache, while a later snapshot generated on main (for example, `0.4.3.9000`) is downloaded and used.

gadenbuie:
I made a few small changes at the end to improve handling of avoiding updating the snapshot or the cache when the snapshot hasn't changed

- Cached pricing is now treated as a complete snapshot instead of being merged with bundled rows.
- Snapshot freshness is tracked with `updated_at` timestamps. Older caches are ignored, and downloads do not replace newer local packaged data.
- The update workflow avoids commits when normalized pricing is unchanged.
- Tests were expanded around schema compatibility, freshness, and cache replacement.
- The user cache location is now documented.

PR #973, opened 2026-05-06
Fix tool calling in Snowflake provider

### LINKED ISSUE #938: ellmer response structure incompatible with querychat - content$type error
###   Description

When using ellmer (connected to Snowflake Cortex AI) as the client for querychat, chat interactions fail with an error suggesting ellmer's response structure doesn't match what querychat expects.

  **Environment:**
  - ellmer: (latest version)
  - querychat: v0.2.0.9000 (dev version via `pak::pak("posit-dev/querychat/pkg-r")`)
  - R version: 4.5
  - Environment: Posit Team Snowflake Native App, Positron within Workbench
  - Snowflake Cortex AI model: `snowflake/claude-sonnet-4-5`

  **Reproducible Example:**
```
  library(shiny)
  library(bslib)
  library(querychat)
  library(ellmer)
  library(DT)
  library(palmerpenguins)

  # Initialize chat with Snowflake Cortex AI
  chat <- chat(
    name = "snowflake/claude-sonnet-4-5",
    system_prompt = "You are a data analysis expert",
    account = Sys.getenv("SNOWFLAKE_ACCOUNT")
  )

  # Test that ellmer chat works standalone
  chat$chat("Hello")  # This works fine

  # Initialize QueryChat with ellmer client
  qc <- QueryChat$new(
    penguins,
    client = chat,
    greeting = "Hello! Ask me questions about the penguins data."
  )

  # Define UI
  ui <- page_sidebar(
    title = "Penguins Data Explorer",
    sidebar = qc$sidebar(),
    card(
      card_header("Data Results"),
      dataTableOutput("table")
    )
  )

  # Define server
  server <- function(input, output, session) {
    qc_vals <- qc$server()

    output$table <- renderDataTable({
      datatable(qc_vals$df(), fillContainer = TRUE)
    })
  }

  shinyApp(ui = ui, server = server)
```

  Behavior:
  1. App launches successfully
  2. Data table displays correctly
  3. Chat UI appears with the greeting message
  4. When attempting to send a message in the chat, the following error occurs:

```
  Error:
  Warning: ERROR: An error occurred in `chat_append_stream(id="querychat_penguins-chat")`
  Caused by error in `if (content$type == "text") ...`:
  ! argument is of length zero

  Warning in append_stream_task$invoke(chat, input$chat_user_input) :
    ERROR: An error occurred when invoking the ExtendedTask.
  Caused by error in `if (content$type == "text") ...`:
  ! argument is of length zero
```

**Additional Context:**
This is for a guide I'm writing that demonstrates natural language data querying using Snowflake Cortex AI in the Posit Team Native App.  Guidance on how to make it work with querychat (or confirmation that this integration isn't currently supported) would be helpful. Thank you!!

cpsievert:
The PR fixes tool calling in `chat_snowflake()`, which is currently completely broken. Any use of tools (including via querychat) errors with:

```
Error in `if (content$type == "text") ...`:
! argument is of length zero
```

**Reprex** (fails on main, passes on this branch):

```r
library(ellmer)

chat <- chat_snowflake(
  system_prompt = "Always use your tools when asked.",
  account = Sys.getenv("SNOWFLAKE_ACCOUNT")
)
chat$register_tool(tool(
  function(x) paste("result:", x * 2),
  name = "double_number",
  description = "Doubles a number",
  arguments = list(x = type_number("The number to double"))
))

chat$chat("Double the number 5")
```

**Root cause:** Snowflake's streaming API can send `tool_use` as the very first chunk (with no preceding text). The old `stream_merge_chunks` assumed responses always started with text. When they didn't, `content_list` entries ended up without a `type` field, which crashed `value_turn()`.

We didn't catch this because `test_tools_simple(chat_snowflake)` requires live credentials and is always skipped in CI (via `default_snowflake_credentials()` → `testthat::skip()` at `R/provider-snowflake.R:470`).

**Fixes (each a separate commit for clarity):**

1. **Fix `stream_merge_chunks`** — Handle tool_use-first responses; use positional search instead of hardcoded indices; normalize `content_list` entries with explicit `type` fields.
2. **Add `stream_content` override** — Snowflake uses `delta$text`, not `delta$content`. Without this, `stream = "content"` mode (used by querychat) never yielded text during streaming.
3. **Handle tool-only turns in `as_json`** — Once tools parse correctly, the follow-up request failed because serializing a `ContentToolRequest`-first turn was unhandled.
4. **Filter empty text in `value_turn`** — Snowflake sends a final empty text delta; if included in the Turn and sent back as history, the API rejects it with "text content blocks must be non-empty".

Closes #938

## Test plan

- [x] Existing offline chunk-merge test updated for new format
- [x] New offline test for tool_use-first chunk merging (the #938 scenario)
- [x] New offline test for `stream_content` extraction from Snowflake's delta format
- [x] New offline test for `as_json` with tool-only assistant turns
- [x] Verified end-to-end against live Snowflake API (non-streaming + streaming + content mode)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,7 @@
 * `chat_ollama()` now supports `params(top_k = )` (@frankiethull, #896).
 * `chat_openai()` no longer fails when streaming web search results for `web_search_call` action types other than `search` (e.g. `open_page`, `find_in_page`) (#941).
 * `chat_openai()` now uses the default prices if the service tier is missing (@trangdata, #903).
+* `chat_snowflake()` now correctly handles tool calling. Previously, when Snowflake's streaming API sent a tool-use chunk as the very first response (with no preceding text), the chunk merging logic produced malformed content, causing "argument is of length zero" errors (#938).
 * `default_google_credentials()` no longer skips application default credentials (e.g. `GOOGLE_APPLICATION_CREDENTIALS`) in interactive sessions, instead falling through to the OAuth browser flow only when no gargle token is available (@stefanlinner, #922).
 * `models_anthropic()` (and `models_claude()`) gains a `credentials` argument for consistency with `chat_anthropic()` and other `models_*()` functions (@jcrodriguez1989, #917).
 * New `stream_controller()` enables programmatic cancellation of streaming chat responses, e.g. from a Shiny "Cancel" button with `chat$stream()` or `chat$stream_async()`. Streaming turns are now saved incrementally so that partial responses survive cancellation, interrupts (Ctrl-C), and errors. Incomplete turns are recorded as `AssistantPartialTurn` objects, display as interrupted in the chat history, and are included in subsequent model context like complete turns (#643).
--- a/R/provider-snowflake.R
+++ b/R/provider-snowflake.R
@@ -323,6 +323,9 @@ method(as_json, list(ProviderSnowflakeCortex, Turn)) <- function(
     # tool result in textual format here, too -- otherwise it gets confused,
     # like it can't see the output.
     content <- tool_string(x@contents[[1]])
+  } else if (S7_inherits(x@contents[[1]], ContentToolRequest)) {
+    # Tool-only response (no preceding text).
+    content <- "<empty>"
   } else {
     cli::cli_abort("Unsupported content type: {.cls {class(x@contents[[1]])}}.")
   }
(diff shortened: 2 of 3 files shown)


hadley:
I assume you tested this live?

cpsievert:
Yep, also tested with the app from #938

Issue #974, opened 2026-05-06
Streaming ContentThinking chunks lack proper <thinking> tag boundaries


cpsievert:
## Current behavior

When streaming with `stream = "content"`, each `ContentThinking` chunk is yielded as an independent object. If you print or concatenate these (via `format()`), each fragment gets its own `<thinking>...</thinking>` wrapper:

```r
chat <- chat_anthropic(model = "claude-sonnet-4-20250514")

stream <- chat$stream("Explain recursion", stream = "content")
coro::loop(for (chunk in stream) {
  cat(format(chunk))
})
```

Output:

```
<thinking>
The user wants an explanation
</thinking>
<thinking>
 of recursion. I should
</thinking>
<thinking>
 start with a simple definition...
</thinking>
Recursion is a technique where...
```

With `stream = "text"`, thinking text is yielded as bare strings indistinguishable from response text:

```r
stream <- chat$stream("Explain recursion", stream = "text")
coro::loop(for (chunk in stream) {
  cat(chunk)
})
```

Output:

```
The user wants an explanation of recursion. I should start with a simple definition...Recursion is a technique where...
```

No way to tell where thinking ended and the response began.

## Expected behavior

The streaming layer should emit `<thinking>` tags at the boundaries — once at the start of thinking, once at the end — so that concatenated output is well-formed:

For `stream = "text"`:
```
<thinking>
The user wants an explanation of recursion. I should start with a simple definition...
</thinking>

Recursion is a technique where...
```

For `stream = "content"`, the yielded objects stay the same (`ContentThinking` per chunk), but the echo/display output should show proper boundaries.

## Why this matters

Downstream consumers (like shinychat) that receive a stream and need to separate thinking from response content currently have to implement their own stateful tracking of `ContentThinking` objects, accumulate the thinking text, and reconstruct the tag boundaries themselves. If the stream already had correct boundaries in text mode, consumers could treat it as a text stream with well-formed `<thinking>` tags and parse accordingly — no type inspection needed.

gadenbuie:
Personally, I think that the current behavior of not emitting thinking tokens when `stream = "text"` is correct. I think that in this case, we should only stream the assistant message as text. Note that we don't currently attempt to serialize tool calls or results to text in this case.

[...]

PR #975, opened 2026-05-06
Introduce ContentThinkingDelta for streaming thinking

cpsievert:
## Summary

Addresses #974 by introducing a `ContentThinkingDelta` class that cleanly separates streaming thinking fragments from complete `ContentThinking` objects stored in turn history.

- **`stream = "text"`**: Thinking is suppressed from yielded chunks entirely (consumers only see response text)
- **`stream = "content"`**: Thinking fragments are yielded as typed `ContentThinkingDelta` objects with a `phase` property (`"start"`, `"body"`, or `"end"`) that communicates block boundaries
- **Console echo**: Thinking is always echoed wrapped in `<thinking>` / `</thinking>` tags so users can distinguish it from response text

### `$chat()` — thinking echoed with tags

```r
chat <- chat_anthropic(
  model = "claude-sonnet-4-20250514",
  params = params(budget_tokens = 1024, max_tokens = 2048)
)

chat$chat("What is 2+2? Think carefully.")
```

Console output:

```
<thinking>
This is a very basic arithmetic question. 2 + 2 = 4. The human is asking me to
"think carefully" but this is straightforward addition. There's no trick here -
it's just basic math.
</thinking>

2 + 2 = 4

This is a fundamental addition problem. When you add 2 and 2 together, you get
4.
```

### `$stream(stream = "text")` — thinking suppressed from yielded chunks

```r
chat <- chat_anthropic(
  model = "claude-sonnet-4-20250514",
  params = params(budget_tokens = 1024, max_tokens = 2048)
)

chunks <- coro::collect(chat$stream("What is 2+2? Think carefully.", stream = "text"))
cat(paste0(chunks, collapse = ""))
```

Output:

```
2 + 2 = 4

I thought carefully about this, and it's straightforward addition. Two plus two
equals four.
```

## Design decisions

- **Two classes**: `ContentThinking` (complete thought, stored in turns, `format()` wraps in tags) vs `ContentThinkingDelta` (streaming fragment, never stored in turns, `format()` returns bare text)
- **Providers emit `ContentThinkingDelta`**: Providers stay simple — they just emit deltas with the default `phase = "body"`. The central streaming layer annotates phase based on transitions.
- **Sentinel end marker**: The `phase = "end"` delta has empty thinking text and signals "thinking has ended" without requiring look-ahead buffering.
- **No tag injection into the stream**: Typed objects carry boundary information — consumers decide how to render. This addresses @gadenbuie's feedback that objects should carry structure, not strings.

## What downstream consumers (shinychat) get

```
ContentThinkingDelta("first reasoning...", phase = "start")
ContentThinkingDelta("more reasoning...", phase = "body")
ContentThinkingDelta("", phase = "end")
ContentText("Here is my response...")
```

Shinychat can:
1. Route thinking deltas to the thinking panel with `content_type: "thinking"` wire flag
2. Use `phase` to manage stored-message boundaries (prepend `<thinking>\n` on start, append `\n</thinking>\n\n` on end)
3. Keep its existing client-side tag parser for raw-tag models (DeepSeek, QwQ)

## Test plan

- [x] `stream = "text"` suppresses thinking from yielded output
- [x] `stream = "content"` yields `ContentThinkingDelta` objects with correct phase sequence (start → body → end)
- [x] Thinking chunks precede text chunks in stream order
- [x] All existing tests pass


cpsievert:
@gadenbuie ready for another look

PR #979, opened 2026-05-12
fix: when using batch_chat and batch_chat_structured, return `NULL` when job incomplete and `wait=FALSE`

### LINKED ISSUE #891: batch_chat_structured <> wait = FALSE doesnt work in ellmer 0.4.0
Hi 👋

The argument wait = FALSE in `batch_chat_structured `no longer works since version 0.4.0.
It now throws the following error:

Error in job$result_turns():
! Provider returned unexpected number of responses.
✖ Expected 2, got 0.

This used to work fine in version 0.3.0 (the call would return immediately while the job ran asynchronously).

Here’s a minimal reprex:

```
library(ellmer)

chat <- chat_openai(model = "gpt-4.1")

prompts <- list(
  list(
    "What do you see in this image?",
    content_image_url("https://www.r-project.org/Rlogo.png")
  ),
  list(
    "What do you see in this image?",
    content_image_file(system.file("httr2.png", package = "ellmer"))
  )
)

data <- batch_chat_structured(
  chat = chat,
  prompts = prompts,
  path = "imageC-data.json",
  type = ellmer::type_string(
    description = "Description of the provided image."
  ),
  wait = FALSE
)

```
Error output:

Error in job$result_turns():
! Provider returned unexpected number of responses.
✖ Expected 2, got 0.
Run rlang::last_trace() to see where the error occurred.

thisisnic:
Fixes #891

This PR adds an early return of `NULL` when a job is not completed, as described in the docs.

### DIFF
--- a/R/batch-chat.R
+++ b/R/batch-chat.R
@@ -89,9 +89,8 @@ batch_chat <- function(chat, prompts, path, wait = TRUE, ignore_hash = FALSE) {
     wait = wait,
     ignore_hash = ignore_hash
   )
-  result <- job$step_until_done()
-  if (is.null(result)) {
-    return(NULL)
+  if (is.null(job$step_until_done())) {
+    return()
   }
 
   assistant_turns <- job$result_turns()
@@ -153,9 +152,8 @@ batch_chat_structured <- function(
     wait = wait,
     ignore_hash = ignore_hash
   )
-  result <- job$step_until_done()
-  if (is.null(result)) {
-    return(NULL)
+  if (is.null(job$step_until_done())) {
+    return()
   }
   turns <- job$result_turns()
 
--- a/R/batch-chat.R
+++ b/R/batch-chat.R
@@ -89,7 +89,10 @@ batch_chat <- function(chat, prompts, path, wait = TRUE, ignore_hash = FALSE) {
     wait = wait,
     ignore_hash = ignore_hash
   )
-  job$step_until_done()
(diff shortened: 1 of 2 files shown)


R/batch-chat.R
```diff
@@ -89,7 +89,10 @@ batch_chat <- function(chat, prompts, path, wait = TRUE, ignore_hash = FALSE) {
     wait = wait,
     ignore_hash = ignore_hash
   )
-  job$step_until_done()
+  result <- job$step_until_done()
+  if (is.null(result)) {
+    return(NULL)
+  }
```
hadley: I'd write this as:

```suggestion
  if (is.null(job$step_until_done())) {
    return()
  }
```

since we don't use `result` later, and the default value of `return()` is `NULL`.

tests/testthat/test-batch-chat.R
```diff
@@ -144,7 +144,42 @@ test_that("errors if wait = FALSE and not complete", {
...
+  result <- batch_chat(
+    chat_openai_test(),
+    list("What's your name"),
+    path = path,
+    wait = FALSE
+  )
+  expect_null(result)
+})
+
+test_that("batch_chat_structured returns NULL when wait = FALSE and not complete", {
```
hadley: Since so much of this is similar to the test above, I'd combine them into a single test.

hadley:
Thanks!

Issue #931, opened 2026-02-19
Tool call escapes the JSON string


dipterix:
_The following example tries to provide full context demonstrating the issues. The bug not always shows due to the unexpected nature of AI agent._

  Reproduces a bug where qwen3-coder:30b on Ollama wraps every tool string
  argument in JSON array brackets:

    Expected: write_file(path = "SKILL.md", content = "hello")
    Actual:   write_file(path = '["SKILL.md"]', content = '["hello"]')

  This makes all tool calls fail when the tool validates its inputs.

  Root cause: The MODEL generates malformed tool call arguments.
  Ollama's OpenAI-compatible API faithfully passes them through,
  and ellmer's ProviderOpenAICompatible does not coerce array→string.

  Question: should ellmer unwrap single-element arrays when the
  tool schema declares type_string()? Although `'["SKILL.md"]'` is also a string, the model may go crazy into escaping strings (sometimes `'\\\\[\\\\\\\\"SKILL.md\\\\\\\\"\\\\]'` after several failed attempts.)

  Prerequisites:
    - Ollama running at the URL below
    - Model: ollama pull qwen3-coder:30b

``` r
# ====================================================================
# test-ollama-array-wrap-bug.R
# ====================================================================


library(ellmer)

ollama_url <- "http://127.0.0.1:11434"

chat <- chat_ollama(
  base_url = ollama_url,
  model = "qwen3-coder:30b",
  system_prompt = paste(
    "You are a file-writing assistant.",
    "When asked to create a file, use the write_file tool.",
    "Do NOT ask questions. Just write the file.",
    "IMPORTANT! DO NOT CALL TOOLS IN PARALLEL; ALWAYS CALL THEM IN SEQUENTIAL ORDER: THE PLATFORM YOU ARE RUNNING ON CRASHES/SEGFAULTS WITH PARALLEL TOOL CALLS.",
    "/nothink"
  )
)

# --- Tool that logs what it receives ---
received_args <- list()

write_file_tool <- tool(
  function(path, content) {
    received_args$path <<- path
    received_args$content <<- content
    cat("\n=== write_file CALLED ===\n")
    cat("  path type:  ", typeof(path), "\n")
    cat("  path value: ", repr::repr(path), "\n")
    if (grepl("^\\[", path)) {
      stop(sprintf("BUG: path wrapped in array brackets: %s", path))
    }
    paste0("File written: ", path)
  },
  name = "write_file",
[...]

PR #980, opened 2026-05-13
Defensively unwrap elements when converting R objects to scalar types and input is length-1 list

thisisnic:
In #931, there is an error where models sometimes wrap tool call arguments in JSON array brackets even though the schema declares a scalar type, which causes the tool to fail. 

Although this is a model error, we can add in a check that when we know the expected type is scalar, then if a length 1 list is passed in, extract the first element.

### DIFF
--- a/R/chat-structured.R
+++ b/R/chat-structured.R
@@ -81,6 +81,8 @@ convert_from_type <- function(x, type) {
         string = NA_character_,
         cli::cli_abort("Unknown type {type@type}", .internal = TRUE)
       )
+    } else if (is.list(x) && length(x) == 1) {
+      x[[1]]
     } else {
       x
     }
--- a/R/chat-structured.R
+++ b/R/chat-structured.R
@@ -82,6 +82,7 @@ convert_from_type <- function(x, type) {
         cli::cli_abort("Unknown type {type@type}", .internal = TRUE)
       )
     } else if (is.list(x) && length(x) == 1) {
+      # Some models intermittently wrap scalars in a JSON array (#931)
       x[[1]]
     } else {
       x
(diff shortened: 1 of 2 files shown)


tests/testthat/test-chat-structured.R
```diff
@@ -65,6 +65,13 @@ test_that("optional base types (scalars) stay as NULL", {
   expect_equal(convert_from_type(NULL, type_string(required = FALSE)), NULL)
 })
 
+test_that("single-element arrays are unwrapped for scalar types", {
```
hadley:
```suggestion
test_that("single-element arrays are unwrapped for scalar types (#931)", {
```

R/chat-structured.R
```diff
@@ -81,6 +81,8 @@ convert_from_type <- function(x, type) {
         string = NA_character_,
         cli::cli_abort("Unknown type {type@type}", .internal = TRUE)
       )
+    } else if (is.list(x) && length(x) == 1) {
+      x[[1]]
```
hadley, line 86: Your reasoning makes sense to me, but I'd also include it here as a brief comment so we don't get confused when looking at in the future.

hadley (commented):
Does this fix the underlying issue?

R/chat-structured.R
thisisnic: Done!

thisisnic:
> Does this fix the underlying issue?

I guess technically (if I'm being pedantic), nope, as it looks like it's an issue with the model itself (or ollama) and not ellmer, but it should prevent it from causing problems.

hadley (approved):
Feel free to merge yourself now 😄 

(And I do think you should close the attached issue)

PR #981, opened 2026-05-13
Enable reasoning effort parameters in `chat_ollama()` and parse thinking contents from `reasoning` field

### LINKED ISSUE #940: Not possible to pass `think = FALSE` to Ollama models via `chat_ollama()`
I’m using `chat_ollama()` with a local Ollama model (`qwen3.5:4b`). Ollama allows to [disable reasoning](https://ollama.com/blog/thinking) for thinking models using the `think = false` parameter (CLI flag `--think=false`). The provided Python code works. However, I haven’t found a way to pass this parameter through `chat_ollama()` with ellmer. 

From looking at the implementation, it seems that:

* `chat_ollama()` uses the OpenAI-compatible API (`/v1/chat/completions`).
* The `chat_params()` method for `ProviderOllama` maps only a limited set of parameters (`temperature`, `top_p`, `max_tokens`, etc.):

```r
method(chat_params, ProviderOllama) <- function(provider, params) {
  standardise_params(
    params,
    c(
      frequency_penalty = "frequency_penalty",
      presence_penalty = "presence_penalty",
      seed = "seed",
      stop = "stop_sequences",
      temperature = "temperature",
      top_p = "top_p",
      top_k = "top_k",
      max_tokens = "max_tokens"
    )
  )
}
```

I added `max_tokens = "max_tokens"` locally to the package and re-installed it, but that did not change anything. I also tried passing the argument via `api_args`:

```r
chat <- chat_ollama(
  model = "qwen3.5:4b",
  api_args = list(think = FALSE)
)
```

but this also does not seem to propagate the parameter to the Ollama request.

Is there currently a supported way to pass `think = FALSE` to Ollama models when using `chat_ollama()` or is this limitation due to using the OpenAI-compatible endpoint instead of the native Ollama API?

thisisnic:
Fixes #940


thisisnic:
When working on this, Claude identified a follow-up task, but I want to check that the approach on this PR makes sense before opening an issue.

:robot: "**Follow-up needed:** The multi-turn write path in `ProviderOpenAICompatible` (line 405) serializes `ContentThinking` back as `reasoning_content` when sending conversation history. Ollama (and increasingly the broader ecosystem) uses `reasoning` as the field name instead. This means if `preserve_thinking = TRUE` is set with Ollama, the thinking from earlier turns may not be recognized by the API. This doesn't bite anyone right now since `preserve_thinking` defaults to `FALSE` for Ollama, but it's worth addressing as part of the broader `reasoning_content` → `reasoning` migration."

R/provider-openai-compatible.R
```diff
@@ -254,7 +254,7 @@ method(stream_content, ProviderOpenAICompatible) <- function(provider, event) {
   }
   delta <- event$choices[[1]]$delta
 
-  reasoning <- delta[["reasoning_content"]]
+  reasoning <- delta[["reasoning_content"]] %||% delta[["reasoning"]]
```
gadenbuie: Thanks for calling this out in the PR description. I looked into it more and OpenAI's docs are instructive here:

> There is currently no generally agreed upon specification in the community with the general properties on a message being either `reasoning` or `reasoning_content`. **To be compatible with clients like the OpenAI Agents SDK we recommend using a `reasoning` field as the primary property for the raw CoT in Chat Completions**.
>
> **Source:** https://developers.openai.com/cookbook/articles/gpt-oss/verifying-implementations#chat-completions

There's also a similar statement in another article which recommends following OpenRouter's implementation, which states:

> For models that only return raw reasoning strings, you can use the simpler `reasoning` field. You can also use `reasoning_content` as an alias - it functions identically to `reasoning`.
>
> **Source:** https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#preserving-reasoning

So the more externally accurate approach would be

```suggestion
  reasoning <- delta[["reasoning"]] %||% delta[["reasoning_content"]]
```

As for including reasoning in API calls when preserving thinking, when I implemented this I was mostly using local models and testing with LM Studio so I could verify the API calls directly in their console. Unfortunately, LM Studio only uses `reasoning_content` in their API calls. [OpenRouter aliases `reasoning_content`](https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#preserving-reasoning), but I could imagine other implementations using only `reasoning`. I wonder if that points toward needing to include a field on `ProviderOpenAICompatible` to store the field name; or possibly we could overload `preserve_thinking` to take a string for the field where reasoning should be preserved (I'm less into this idea but it's small and scoped and might be preferred over adding another field).

thisisnic: Thanks for the links to those docs, they're super helpful!

> I could imagine other implementations using only `reasoning`

Yeah, it seems less common, but in the [vLLM docs](https://docs.vllm.ai/en/stable/features/reasoning_outputs/) they mention only using `reasoning`. I checked to see if they alias it, but they [removed `reasoning_content` entirely](https://github.com/vllm-project/vllm/pull/33402).

hadley (approved):

PR #976, opened 2026-05-10
Implement official Claude structured outputs

thisisnic:
Implements native structured outputs for Claude.  This is only available for later models, so have retained the "spoof it via a tool call" path for older models.  Added a deprecation warning for `additional_properties` as well.

Closes #866

PR #983, opened 2026-05-14
Update prompt for Claude structured output tool

### LINKED ISSUE #881: `$chat_structured()` fails with claude sometimes
``` r
library(ellmer)
#> Warning: package 'ellmer' was built under R version 4.5.2
chat <- chat_claude(model = "claude-haiku-4-5-20251001")
type_person <- type_object(
  name = type_string(),
  age = type_number()
)
chat$chat_structured("Cheese", type = type_person)
#> Error in x[[name]]: subscript out of bounds
```

<sup>Created on 2025-11-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thisisnic:
Fixes #881 

This fixes the consistent failures show in the linked issue, but is still a bit messy, i.e. we end up with a character with value "<UNKNOWN>" where we requested a numeric return value for `age`.

I was thinking of trying to do something about the typing issues, but I'm concerned about overcomplicating the prompt and potentially increasing the chance of more inconsistencies.  Happy to take this on if we think it's worth it, though I'm not convinced it's worth investing the effort given that this only affects models which can't use the Anthropic API's native structured output functionality implemented in #976, and so will likely phase out over time anyway.

Haven't added unit tests here as this is untested at the moment.

We also do this tool-as-structured-output for AWS Bedrock, and so we should also consider whether we want to update the prompt there (we can factor it out into its own function if it's going to be the same but I don't have any Bedrock credentials yet so can't test out if it has the same issues) and I also will open an issue around considering if we want to use native structured output for Bedrock models which now support it.

Old behaviour:

``` r
library(ellmer)
chat <- chat_claude(model = "claude-sonnet-4-20250514")
type_person <- type_object(
  name = type_string(),
  age = type_number()
)
chat$chat_structured("Cheese", type = type_person)
#> Error in x[[name]]: subscript out of bounds
```

New behaviour:

``` r
library(ellmer)
type_person <- type_object(
  name = type_string(),
  age = type_number()
)

# Native structured output
chat_so <- chat_claude(model = "claude-sonnet-4-5")
chat_so$chat_structured("Cheese", type = type_person)
#> $name
#> [1] "Cheese"
#> 
#> $age
#> [1] 0


# Tool-based structured output
chat_tool <- chat_claude(model = "claude-sonnet-4-20250514")
chat_tool$chat_structured("Cheese", type = type_person)
#> $name
#> [1] "<UNKNOWN>"
#> 
#> $age
#> [1] "<UNKNOWN>"
```

<sup>Created on 2026-05-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/chat-structured.R
+++ b/R/chat-structured.R
@@ -35,6 +35,15 @@ wrap_type_if_needed <- function(type, needs_wrapper = FALSE) {
   }
 }
 
+structured_tool_def <- function(name, type) {
+  ToolDef(
+    function(...) {},
+    name = name,
+    description = "Extract structured data. Return a valid JSON object conforming exactly to the provided schema with all required fields.",
+    arguments = type_object(data = type)
+  )
+}
+
 convert_from_type <- function(x, type) {
   if (is.null(x) && !type@required) {
     x
--- a/R/chat-structured.R
+++ b/R/chat-structured.R
@@ -35,15 +35,6 @@ wrap_type_if_needed <- function(type, needs_wrapper = FALSE) {
   }
 }
 
-structured_tool_def <- function(name, type) {
-  ToolDef(
-    function(...) {},
-    name = name,
-    description = "Extract structured data. Return a valid JSON object conforming exactly to the provided schema with all required fields.",
-    arguments = type_object(data = type)
-  )
-}
-
 convert_from_type <- function(x, type) {
   if (is.null(x) && !type@required) {
     x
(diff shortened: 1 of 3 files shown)


thisisnic:
Hmm, this might be pointless - upon more testing, it seems that the older Sonnet models still seem to get the right answer without the fix, even when calling a tool for structured output, and the Haiku model used in the original issue now uses the native structured output path.  I will check out if there are any available models which would get it wrong, and just abandon this PR if there are none.

PR #985, opened 2026-05-18
Handle mcp_tool_use and mcp_tool_result content types

klin333:
Completely vibe coded by Claude. But seems to do the job. I have absolutely zero experience or expertise in this package or field. 

essentially is to handle mcp on claude api: https://platform.claude.com/docs/en/agents-and-tools/mcp-connector


hadley:
@gadenbuie I think you're the expert here 😄

PR #992, opened 2026-05-20
Add `set_model()` method to `Chat`

### LINKED ISSUE #988: Make it easier to switch models from `Chat` or `Provider` objects
It's currently quite difficult to take a `Chat` or `Provider` object and create a new conversation with a different model. Related to #499 because currently `Provider` stores the `model`.

Ideally, changing models would be as easy as `chat$set_model("new model")`

gadenbuie:
Closes #988

## Summary

Adds a `set_model()` method to the `Chat` R6 class, complementing the existing `get_model()`. This lets users switch models mid-conversation without recreating the chat object.

Unlike the `chat_*()` constructors, `set_model()` does not validate the model name against available models for the provider — it accepts any string.

## Verification

```r
chat <- chat_openai(model = "gpt-5.3")
chat$chat("Tell me a joke about R programming")

# Switch to a different model for the next turn
chat$set_model("gpt-5.4")
chat$get_model()
#> [1] "gpt-5.4"

chat$chat("Now tell me a joke about Python")
```

### DIFF
--- a/R/chat.R
+++ b/R/chat.R
@@ -100,6 +100,16 @@ Chat <- R6::R6Class(
       private$provider@model
     },
 
+    #' @description Update the model name. Note that unlike some of the
+    #'   `chat_*()` functions, the model name is not validated against available
+    #'   models for the provider.
+    #' @param model A single string giving the new model name.
+    set_model = function(model) {
+      check_string(model)
+      private$provider@model <- model
+      invisible(self)
+    },
+
     #' @description Update the system prompt
     #' @param value A character vector giving the new system prompt
     set_system_prompt = function(value) {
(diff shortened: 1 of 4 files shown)


R/chat.R
```diff
@@ -100,6 +100,16 @@ Chat <- R6::R6Class(
       private$provider@model
     },
 
+    #' @description Update the model name. Note that unlike some of the
+    #'   `chat_*()` functions, the model name is not validated against available
+    #'   models for the provider.
```
gadenbuie, line 105: This is one way that Provider and Chat could be linked -- i.e. for ollama and similar we'd probably want to validate the model name like we do in `chat_ollama()`. I'm punting on this for now to leave us room to decide in the `Provider` redesign what we want to do here.

hadley (approved):
Oh yes this is simple 😆

PR #996, opened 2026-05-22
Clean up Google batch tests

### LINKED ISSUE #995: Clean up tests for Google Gemini Batch support
I reviewed and merged #926 but missed that the tests aren't quite aligned with tests for other providers - it's probably worth trimming these down a bit and updating them.

thisisnic:
I accidentally overlooked the tests when reviewing #926 as they were hidden by the GitHub UI, and it looks like the original Claude-generated code has done a little bit of LLM overtesting - I've refactored the tests to be more in line with the other providers' ones and tweaked the naming of a couple of other tests for clarity.

I left a couple in that are specifically about handling Google Gemini API quirks.

I've also updated the snapshot test for Google as they updated their API recently.

Closes #995


tests/testthat/test-provider-claude.R
```diff
@@ -1,5 +1,5 @@
-test_that("can make simple batch request", {
-  vcr::local_cassette("anthropic-batch")
+test_that("can make simple request", {
```
hadley, line 1: Oops

hadley (approved):
Looks good!

PR #1005, opened 2026-05-30
docs: add OpenAI-compatible gateway example

cerebrixos:
## Summary
- add a Tuning Engines example for chat_openai_compatible()
- show base_url, API key, and model setup in README.Rmd and generated README.md
- keep the change in the existing provider overview section

## Why
We are an AI infrastructure team using ellmer with Tuning Engines. Since ellmer already supports OpenAI-compatible providers and gateways through chat_openai_compatible(), this gives R users a concrete setup for governed model access, policy checks, traces, audit logs, and usage/cost reporting. It would mean a lot to us if this helps your users, and I am happy to adjust wording or placement.

## Validation
- git diff --check

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -61,6 +61,20 @@ ellmer supports a wide variety of model providers:
 * Snowflake Cortex: `chat_snowflake()` and `chat_cortex_analyst()`.
 * VLLM: `chat_vllm()`.
 
+You can also use OpenAI-compatible gateways with `chat_openai_compatible()`.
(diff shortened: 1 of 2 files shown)


thisisnic:
Thanks for the PR! 

We've already document OpenAI-compatible endpoints at [`chat_openai_compatible()`](https://ellmer.tidyverse.org/reference/chat_openai_compatible.html), which would cover Tuning Engines, and we can't mention every possible provider as there are quite a lot out there.  If Tuning Engines gets significant traction/usage in the community, we'd be happy to look at it again.

PR #1006, opened 2026-05-31
default base_url to OLLAMA_BASE_URL for all ollama funcs

debruine:
Set default for base_url to Sys.getenv("OLLAMA_BASE_URL", "http://localhost:11434") instead of just "http://localhost:11434" in models_ollama() and has_ollama() so they match the default value in chat_ollama()

### DIFF
--- a/R/provider-ollama.R
+++ b/R/provider-ollama.R
@@ -164,7 +164,7 @@ skip_if_no_ollama <- function() {
 #' @export
 #' @rdname chat_ollama
 models_ollama <- function(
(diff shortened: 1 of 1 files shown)


debruine:
Ignore me, I see now the credentials argument does this. It's still confusing that the default base_url is different between chat_ollama and models_ollama, but my edit isn't working.

thisisnic:
@debruine I noticed this closed PR, and I think you're actually right here...happy to review it if you want to reopen it?

PR #993, opened 2026-05-21
feat: add `models_list()` S7 generic to Provider class

gadenbuie:
Closes #986

## Summary

Adds an internal S7 generic `models_list(provider)` that returns a data frame of available models for a given provider. Each provider's `models_*()` standalone function is refactored to delegate to `models_list()`, keeping the public API unchanged.

This lays the groundwork for a future Provider redesign (#499) where `models_list()` could become exported and user-facing. Advanced users and risk-takers can use `ellmer:::models_list()` in the meantime.

**Key design decisions:**

- `models_list()` is internal (non-exported). The base `Provider` method throws a classed `"not_implemented"` error; `ProviderOpenAICompatible` provides a working default for the standard `/models` endpoint.
- A `Chat` dispatch method allows `ellmer:::models_list(chat)` as a convenience.
- Each provider method handles its own URL transformations (e.g., Ollama strips `/v1`, AWS swaps `bedrock-runtime` to `bedrock`) so `models_*()` wrappers stay thin.
- Providers keep their existing return columns -- no standardization across providers.
- A new `ProviderGitHub` class was introduced (empty subclass of `ProviderOpenAICompatible`) so that `chat_github() |> models_list()` correctly dispatches to `/catalog/models` rather than the OpenAI-compatible `/models` default.

**13 methods registered:** base `Provider` (error), `Chat` (delegation), `ProviderOpenAICompatible` (default), plus 10 provider-specific methods for Anthropic, Google (Gemini + Vertex), AWS Bedrock, Ollama, LMStudio, vLLM, DeepSeek, Mistral, GitHub, and Portkey.

## Example

``` r
pkgload::load_all()
#> ℹ Loading ellmer
chat_claude() |> ellmer:::models_list() |> dplyr::as_tibble()
#> Using model = "claude-sonnet-4-5-20250929".
#> # A tibble: 9 × 6
#>   id                         name  created_at          cached_input input output
#>   <chr>                      <chr> <dttm>                     <dbl> <dbl>  <dbl>
#> 1 claude-opus-4-7            Clau… 2026-04-14 00:00:00          0.5     5     25
#> 2 claude-sonnet-4-6          Clau… 2026-02-17 00:00:00          0.3     3     15
#> 3 claude-opus-4-6            Clau… 2026-02-04 00:00:00          0.5     5     25
#> 4 claude-opus-4-5-20251101   Clau… 2025-11-24 00:00:00          0.5     5     25
[...]

PR #1008, opened 2026-06-01
Implement `models_groq()`

### LINKED ISSUE #921: Add models_groq()
This is what we're using in metacheck:

```
#' Get Groq Models
#'
#' Returns a list of active models in groq. See <https://console.groq.com/docs/models> for more information.
#'
#' @returns a table of model info
#' @export
models_groq <- function() {
  API_KEY <- Sys.getenv("GROQ_API_KEY")
  url <- "https://api.groq.com/openai/v1/models"
  config <- httr::add_headers(
    Authorization = paste("Bearer", API_KEY)
  )

  response <- httr::GET(
    url, config,
    encode = "json")

  models <- do.call(dplyr::bind_rows,
                    httr::content(response)$data) |>
    data.frame()

  # fix `created` column format to fit ellmer::models_platform() tables
  models$created_at <- as.POSIXct(models$created) |> format("%Y-%m-%d") |> as.Date()
  rows <- models$active
  cols <- names(models) |> setdiff(c("active", "created"))
  active <- models[rows, cols]

  return(active)
}
```

thisisnic:
Closes #921 

Adds `models_groq()` function.  I sorted by `owned_by` column (i.e. provider) as it felt like a reasonable choice for scannign the table by eye.  Also added to the pricing updates.


gadenbuie:
Hi @thisisnic! I just merged #993 which added a `models_list()` generic for providers and I pulled the changes into this branch. It looks like we can now rely on the default `models_list()` for `ProviderOpenAICompatible`. I don't have a Groq key though, so if you have a way to test this, that would be great. Also this does mean the sorting order will change, but I think that's okay.

thisisnic:
That's great @gadenbuie; yay for less duplication, and thanks for pulling the changes in here!  Have tested locally and all looks good here - agree ordering doesn't matter.  I've also added the `sysdata.rda` file that was generated, as I originally had concerns about merge conflicts, but I think it should be fine actually.

hadley (approved):

Issue #867, opened 2025-11-15
chat_structured() silently fails when response hits max_tokens limit


n8layman:
When `chat_structured()` exceeds `max_tokens`, it returns empty/incorrect values with no warning or error. I ran into this trying to extract large bibliographies from scientific articles using claude-sonnet-4-5 with the default output size of 4096 tokens. The whole output schema was NULL. A warning to bump up output context size using params would help users understand what's going on here!

## Reproducible Example

Created with `reprex::reprex(session_info = TRUE)`:

```r
library(ellmer)

# Schema with array field
schema <- ellmer::type_object(
  sentences = ellmer::type_array(
    items = ellmer::type_string(),
    description = "Array of all sentences"
  )
)

# Use low max_tokens to force truncation
chat <- ellmer::chat(
  name = "anthropic/claude-sonnet-4-5",
  system_prompt = "You are a helpful assistant.",
  params = list(max_tokens = 100)
)

large_context <- paste(rep("Lorem ipsum dolor sit amet. ", 500), collapse = "")
result <- chat$chat_structured(large_context, type = schema)

# Check response
turns <- chat$get_turns()
last_turn <- turns[[length(turns)]]

last_turn@json$stop_reason         # "max_tokens"
last_turn@json$usage$output_tokens # 100
class(result$sentences)            # "character" (should be "list")
result$sentences                   # "" (empty string)
```

<sup>Created on YYYY-MM-DD with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;"><summary>Session info</summary>

```
─ Session info ───────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.4.x
 os       macOS
 system   x86_64, darwin
 ellmer   0.3.2
─────────────────────────────────────────────────────────────────────────────
```

</details>

## Observed Behavior

1. API returns `stop_reason: "max_tokens"`
2. Tool input is empty: `{}`
3. `chat_structured()` returns result with empty/default values
4. No warning or error is raised
5. Type mismatch: array field returns as character instead of list

## Expected Behavior

When `stop_reason == "max_tokens"`:

1. Throw error: "Response exceeded max_tokens limit. Increase max_tokens or simplify schema."
2. OR return partial results with warning
3. OR attempt to repair truncated JSON and return partial results

```r
reprex::reprex(session_info = TRUE)
#> ℹ Non-interactive session, setting `html_preview = FALSE`.
[...]

PR #1015, opened 2026-06-10
Add `finish_reason` property to `AssistantTurn`

### LINKED ISSUE #3: Make use of `finish_reason`
https://platform.openai.com/docs/guides/function-calling/edge-cases

Especially the `finish_reason == "length"` case, where things like tool calls might be malformed because the context window was exhausted.

thisisnic:
Closes #3 - and should set things up nicely for #867

I went with collapsing common options a smaller subset so we can use those values to output error messages - planning to implement that in a follow-up PR

## Summary

- `AssistantTurn` gains a `finish_reason` property that standardizes provider-specific stop reasons into a common set: `"success"`, `"max_tokens"`, `"stop_sequence"`, `"content_filter"`, `"context_window"`. Unrecognized reasons are passed through as the raw provider string. `NA_character_` is the default when no provider method exists.
- Finish reason extraction is handled inline in each provider's `value_turn()` method, following the same pattern as `value_tokens()` and `get_token_cost()`.
- Providers implemented: Claude, OpenAI, OpenAI-compatible, Google Gemini, and AWS Bedrock.
- OpenAI streaming fix: added `response.incomplete` handler in `stream_merge_chunks` so truncated streaming responses are returned rather than silently dropped as `NULL`.

## Design decisions

- **Raw pass-through for unknown values:** Unrecognized stop reasons are returned as-is from the provider rather than mapped to `"other"`, so we have the option to add details to error messages without users needing to dig into `turn@json`.
- **`tool_use` excluded** from the standardized set because ellmer handles tool calls internally -- the user never sees those turns.

### DIFF
--- a/R/provider-aws.R
+++ b/R/provider-aws.R
@@ -392,7 +392,8 @@ method(stream_merge_chunks, ProviderAWSBedrock) <- function(
     result <- list(
       output = list(
         message = result
-      )
+      ),
+      stopReason = chunk$stopReason
     )
   } else if (chunk$event_type == "metadata") {
     result$usage <- chunk$usage
--- a/R/provider-aws.R
+++ b/R/provider-aws.R
@@ -413,20 +413,6 @@ method(value_tokens, ProviderAWSBedrock) <- function(provider, json) {
   )
 }
 
-# https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html
-method(value_finish_reason, ProviderAWSBedrock) <- function(provider, json) {
-  switch(
-    json$stopReason,
-    end_turn = "success",
-    max_tokens = "max_tokens",
-    model_context_window_exceeded = "context_window",
-    stop_sequence = "stop_sequence",
-    guardrail_intervened = ,
-    content_filtered = "content_filter",
-    "other"
-  )
-}
-
 method(value_turn, ProviderAWSBedrock) <- function(
   provider,
   result,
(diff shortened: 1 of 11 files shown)


R/provider-aws.R
```diff
@@ -450,7 +451,26 @@ method(value_turn, ProviderAWSBedrock) <- function(
...
+  # https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html
+  finish_reason <- switch(
+    result$stopReason,
+    end_turn = "success",
+    max_tokens = "max_tokens",
+    model_context_window_exceeded = "context_window",
+    stop_sequence = "stop_sequence",
+    guardrail_intervened = ,
+    content_filtered = "content_filter",
+    result$stopReason
```
hadley: I'd wrap this in `I()` just so there's a way to distinguish between standard and not.

R/provider-aws.R
```diff
@@ -450,7 +451,26 @@ method(value_turn, ProviderAWSBedrock) <- function(
 
   tokens <- value_tokens(provider, result)
   cost <- get_token_cost(provider, tokens)
-  AssistantTurn(contents, json = result, tokens = unlist(tokens), cost = cost)
+
+  # https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html
+  finish_reason <- switch(
```
hadley: Maybe worth considering pulling this out into a generic? Like the parameter standardisation stuff.

hadley (approved):
Feel free to merge once you address those two comments

R/provider-aws.R
thisisnic: Done!

PR #1033, opened 2026-06-24
Expose citation metadata from web search tools

### LINKED ISSUE #775: Add content type for citations
https://inspect.aisi.org.uk/reference/inspect_ai.model.html#citation

thisisnic:
Still a draft as I need to do a proper review of the AI-generated code as there's repetition and ugliness in places.

## Summary

- Adds `Citation` S7 class with `url` and `title` properties, and a `citations` property on `ContentText` and `Turn`
- Extracts citations automatically from Claude, OpenAI, and Google Gemini web search responses
- Fixes streaming for Google grounding metadata (was silently dropped)

## Design decisions

- **Citation is not a Content subclass** -- citations are metadata about text, not standalone content. They don't participate in rendering or get sent back to the model.
- **Only `url` and `title`** -- other provider-specific fields (cited_text, start/end indices) differ in semantics across providers and are omitted. Users can access `turn@json` for raw details.
- **Per-provider extraction helpers rather than a generic** -- the three providers receive citations in completely different shapes and at different levels of the response (Claude: per text block, OpenAI: per text part with mixed annotation types, Google: response-level grounding metadata). A shared generic would need an awkward common interface.
- **Google citations attach to first ContentText** -- since grounding metadata is response-level, not per-part. The single-text-part case is overwhelmingly common.

## Usage

``` r
devtools::load_all("~/ellmer")
#> ℹ Loading ellmer
chat <- chat_claude()
#> Using model = "claude-sonnet-4-5-20250929".
chat$chat("When was R 4.4.0 released?")
#> R 4.4.0 ("Puppy Cup") was released on the 24th April 2024.
chat$last_turn()@citations
#> [[1]]
#> What’s new in R 4.4.0? | R-bloggers (<https://www.r-bloggers.com/2024/04/whats-new-in-r-4-4-0/>)
```

Closes #775


R/provider-claude.R
```diff
@@ -794,6 +794,12 @@ cache_control <- function(provider) {
   }
 }
 
+claude_extract_citations <- function(content) {
+  lapply(content$citations %||% list(), function(x) {
+    Citation(url = x$url %||% "", title = x$title %||% "")
+  })
+}
```
thisisnic: I don't love these functions as it feels like they should be generics, but it's a weird case where Claude and OpenAI put citations on each text block, but Google puts them on the response as a whole, so they're inherently different.

R/provider-google.R
```diff
@@ -393,6 +393,17 @@ method(value_turn, ProviderGoogleGemini) <- function(
     }
   })
   contents <- compact(contents)
+
+  citations <- gemini_extract_citations(
+    result$candidates[[1]]$groundingMetadata
+  )
+  if (length(citations) > 0) {
+    idx <- match(TRUE, map_lgl(contents, S7_inherits, ContentText))
```
thisisnic: TODO: update this to have a better variable name than `idx`

cpsievert:
I have some [WIP in chatlas ](https://github.com/posit-dev/chatlas/pull/318) in this direction. It'd be great if you could take that approach into consideration as well. I think what I have is probably a bit more complicated because we want the ability to render citations/footnotes inline in shinychat (i.e., we need `$stream()` to yield citations as they occur).

thisisnic:
> I have some [WIP in chatlas ](https://github.com/posit-dev/chatlas/pull/318) in this direction. It'd be great if you could take that approach into consideration as well. I think what I have is probably a bit more complicated because we want the ability to render citations/footnotes inline in shinychat (i.e., we need `$stream()` to yield citations as they occur).

Sounds good, thanks for the heads up! 

I'll make a note to take a look at existing or in-progress implementations in chatlas when looking at any new ellmer features. How closely are we aiming to couple the implementations?  Are there any particular areas which must/should match between them?

cpsievert:
> How closely are we aiming to couple the implementations? 

Generally speaking, we should be aligned when it comes to headline features or fundamental APIs like streaming content, turns, tools, etc. That said, certain things will make more sense for one language than the other, especially when it comes to API design and implementation (i.e., OOP vs FP). So, in other words, the capabilities should look the same, but the implementations should be free to leverage the strengths/idioms of each respective language. 

> Are there any particular areas which must/should match between them?

I'd mainly view this question through the lens of a user who wishes to implement their project in both packages. If they're able to achieve something in one package and not the other, that's a sign we need to do something differently. We also want to keep in mind that some features (such as this one) may have downstream consequences for projects like shinychat. In that case, it'd be wise to rope either @gadenbuie or I into the loop.

thisisnic:
> > How closely are we aiming to couple the implementations?
> 
> Generally speaking, we should be aligned when it comes to headline features or fundamental APIs like streaming content, turns, tools, etc. That said, certain things will make more sense for one language than the other, especially when it comes to API design and implementation (i.e., OOP vs FP). So, in other words, the capabilities should look the same, but the implementations should be free to leverage the strengths/idioms of each respective language.
> 
> > Are there any particular areas which must/should match between them?
> 
> I'd mainly view this question through the lens of a user who wishes to implement their project in both packages. If they're able to achieve something in one package and not the other, that's a sign we need to do something differently. We also want to keep in mind that some features (such as this one) may have downstream consequences for projects like shinychat. In that case, it'd be wise to rope either @gadenbuie or I into the loop.

Thanks, that clarifies things a lot!  Will keep an eye out for things to ping you both on.

I was going to rewrite this draft PR to follow your chatlas WIP PR, but on reflection, given that you have a pretty clear use case in mind which you're a lot more familiar with than I am, I'm just going to close this for now and revisit it later on, to prevent duplicating effort or needing to change it here if any of the shape of the chatlas PR changes.

Issue #1051, opened 2026-07-04
Feature request: callback before each model request in the tool loop (for mid-loop context compaction)


kaipingyang:
### Summary

`Chat$chat()` / `Chat$chat_async()` (and `stream()/stream_async()`) run the tool loop internally (`chat_impl` / `chat_impl_async` in `R/chat.R`). There is currently **no callback that fires before each model request inside that loop**, so an application cannot inspect or rewrite the conversation **between tool rounds** within a single `chat()` call.

The existing `on_tool_request(callback(request))` and `on_tool_result(callback(result))` hooks are tool-scoped: they receive only the request/result, don't fire when a turn has no tools (or before the first request), and there is no documented/safe contract for mutating `private$.turns` from them.

### Use case: mid-loop context compaction

I'm building an R coding agent on ellmer. To stay within the model's context window I need **automatic compaction between tool rounds** — exactly what Claude Code does by calling `autoCompactIfNeeded()` at the top of each iteration of its query loop. Because the tool loop is internal to ellmer, I can only compact **before** handing off to `chat()` (i.e. at user-turn boundaries). A single turn that triggers many large tool outputs (e.g. reading several big files) can blow the context limit mid-loop, with no opportunity to compact until the turn ends.

### Where it would hook

In `chat_impl` / `chat_impl_async`, at the top of `while (!is.null(user_turn)) { ... }`, before `private$submit_turns(user_turn, ...)`.

### Proposed API (sketch — happy to follow your preferred shape)

A callback fired before each model request in the loop, mirroring the existing `CallbackManager` pattern:

```r
# fires before each submit_turns() inside the tool loop
cancel <- chat$on_turn_start(function() {
  # app can inspect/rewrite history here, e.g. compact:
  if (my_should_compact(chat)) chat$set_turns(my_compact(chat$get_turns()))
})
```

Naming/signature is entirely yours to decide (`on_turn_start` / `on_request` / a `turn_transform` that returns turns, etc.). The key need is a **safe, documented point to rewrite conversation history between tool rounds**.

### Alternatives considered

- `on_tool_result` + a captured `chat` to call `set_turns()`: works only by luck, is undocumented, doesn't fire without tools, and risks orphaning `tool_use`/`tool_result` pairs.
- Re-implementing the tool loop outside ellmer using lower-level primitives: `submit_turns()` is private, so there's no clean public way to drive one model turn + tools manually.

[...]

PR #1052, opened 2026-07-04
feat(Chat): `on_request_start()` and `on_request_end()` hooks

kaipingyang:
### What
Adds two paired `Chat` callbacks fired inside the tool loop, once per model request (whether or not it involves tools):
- **`on_request_start(turns)`** — fires before each request; receives the complete outgoing request (history + the pending turn) for inspection. Compact mid-loop via `chat$set_turns(compact(chat$get_turns()))`.
- **`on_request_end(turn)`** — fires after each response, before its tool calls run; receives the assistant turn (an `AssistantPartialTurn` if the request was cancelled). Useful for latency/cost tracking.

Both are observation-only (like `on_tool_request`/`on_tool_result`); history is rewritten through the existing public `set_turns()` / `get_turns()`.

### Why
Addresses #1051 — there was no hook firing *inside* the tool loop, once per request, independent of tools (needed for mid-loop context compaction).

### Implementation
Two `CallbackManager`s registered in `initialize()`, invoked at the top of the `while` loop and after `last_turn()` in both `chat_impl` / `chat_impl_async`. Tests cover the `set_turns()` rewrite contract and the async loop; NEWS + docs updated. Supersedes the original single `on_turn_start()` per @gadenbuie's review.


hadley:
@gadenbuie can you take a look at this please, since you've spent the most time thinking about callbacks?

At a minimum, I think the docs will need tweaking since I think it uses turn/round in the opposite direction to what we're moving towards.

R/chat.R
```diff
@@ -469,6 +471,43 @@ Chat <- R6::R6Class(
...
+    #'   request** in the tool loop, i.e. immediately before the pending turns
+    #'   are sent to the model. It fires once per request (including the first),
+    #'   whether or not the request involves tools. Use it to inspect or rewrite
+    #'   the conversation before it is sent -- for example, to compact or trim
+    #'   the turns as the conversation approaches the model's context window.
+    #'   Rewriting history is done with `$set_turns()`, which is a supported,
+    #'   documented operation; calling it from this callback is explicitly
+    #'   allowed.
+    #'
+    #' @param callback A function called with a single argument `turns`: the
```
gadenbuie, line 532: This is the spot to nail down the `turns` vs history distinction, since it's the one thing most likely to trip up a callback author. Suggested framing:

- `turns` is the **complete request about to be sent** — the accumulated history *plus* the pending turn (the new user message on the first request, or the synthesized tool-result turn on later ones). It's here so you can *inspect* the full payload (e.g. measure its size, including a large pending tool result) to decide whether to compact.
- To *rewrite* the conversation, use `$set_turns()` / `$get_turns()`, which operate on the **history only** — ellmer always preserves and re-appends the pending turn for you. The canonical idiom is `chat$set_turns(compact(chat$get_turns()))`.
- Worth an explicit warning: **don't pass `turns` straight back into `$set_turns()`.** Because `set_turns()` sets history and ellmer then re-appends the pending turn, `set_turns(turns)` would duplicate the pending turn in the outgoing request.

R/chat.R
```diff
@@ -469,6 +471,43 @@ Chat <- R6::R6Class(
...
+      private$callback_on_request_start$add(callback)
+    },
+
+    #' @description Register a callback that fires at the **end of each model
+    #'   request** in the tool loop, i.e. after the model's response turn is
+    #'   received and before any tool calls it contains are executed. Paired with
+    #'   `$on_request_start()`; useful for per-request latency / cost tracking and
+    #'   for observing tool-call requests before they run.
+    #'
+    #' @param callback A function called with a single argument `turn`: the
```
gadenbuie, line 560: One case worth documenting: when a request is cancelled/interrupted, this still fires (intentionally — so every `on_request_start` has a matching `on_request_end`), but the `turn` may be an `AssistantPartialTurn` with `@tokens` / `@cost` as `NA` and only `@duration` populated. A sentence noting that (and how to detect it, e.g. `turn@reason` or `S7_inherits()`) would save a callback author some confusion when doing cost/latency accounting.

tests/testthat/test-chat-tools.R
```diff
@@ -207,6 +207,28 @@ test_that("chat callbacks for tool requests/results", {
...
   cb_count_request <- 0
   cb_count_result <- 0
+  cb_count_req_start <- 0
+  cb_count_req_end <- 0
+  last_start_turns <- NULL
+  last_end_turn <- NULL
+
+  # Fires before each model request in the tool loop (does not emit, so the
+  # snapshot below is unaffected). Receives the pending turns.
+  chat$on_request_start(function(turns) {
```
gadenbuie, line 217: This asserts the callback fires and inspects `turns`, which is great. Could we also add an assertion for the behavior the feature exists for — that calling `$set_turns()` from *inside* `on_request_start` actually changes what gets sent to the model? Even a small case (e.g. drop or replace a turn in the callback, then assert the outgoing request reflects it) would lock in the compaction contract, which the current call-count/role checks don't cover.

tests/testthat/test-chat-tools.R
```diff
@@ -231,6 +253,14 @@ test_that("chat callbacks for tool requests/results", {
   )
   expect_equal(cb_count_request, 2L)
   expect_equal(cb_count_result, 2L)
+  # One request-start + one request-end per model request: initial request + the
+  # follow-up after the tool results are fed back.
+  expect_equal(cb_count_req_start, 2L)
+  expect_equal(cb_count_req_end, 2L)
```
gadenbuie, line 259: These assertions only exercise the sync `$chat()` loop. Since the callbacks also fire from `chat_impl_async`, a small async test (via `$chat_async()`, and/or `$stream_async()`) would back up the "fires on both sync and async loops" behavior and guard against regressions in the `await(... invoke_async())` path.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # ellmer (development version)
 
+* `Chat` gains `$on_request_start(turns)` and `$on_request_end(turn)`, paired callbacks that fire before and after each model request in the tool loop (once per request, whether or not it involves tools). `$on_request_start()` receives the turns about to be sent, so you can inspect or rewrite the conversation before each request (e.g. compact turns via `$set_turns()` to stay within the context window); `$on_request_end()` receives the assistant turn just returned, useful for per-request latency/cost tracking and observing tool-call requests before they run (#1051).
```
gadenbuie: Small wording tweak to match the final contract: `on_request_start` receives the *complete outgoing request* (history + the pending turn) for inspection, and rewriting the conversation is done with `$set_turns()` (which operates on history). The current "receives the turns about to be sent, so you can inspect or rewrite the conversation" is close but slightly blurs the history-vs-pending line that the docs should make crisp.

gadenbuie (commented):
Thanks for this — and especially for the rework from the original single `on_turn_start()` into the paired `on_request_start(turns)` / `on_request_end(turn)`.

A few things to tighten before merge — almost all of it is documentation and test coverage rather than implementation changes:

1. **Can you please regenerate the docs** with a quick `devtools::document()`? The new `$on_request_start()` / `$on_request_end()` methods don't currently show up in `?Chat` or on the pkgdown site.

2. **Clarify the `turns` contract in the docs** (see inline comment). The mental model worth stating explicitly is that `turns` is the *complete outgoing request* (history **plus** the pending turn), whereas `$get_turns()` / `$set_turns()` operate on *history only*. That distinction deserves to be spelled out.

3. **Can you please update the PR description to match the final implementation?** (It currently describes the original `on_turn_start`.)

None of this is structural — the design is in good shape. Happy to help with any of it.

kaipingyang:
Thanks @gadenbuie — all six addressed in a962c72.

Docs are regenerated (roxygen2 8.0.0, matching `Config/roxygen2/version`), so `$on_request_start()` / `$on_request_end()` now show up in `?Chat` and on the pkgdown reference. The `turns` contract is spelled out on both the method and `@param`: it's the *complete outgoing request* (history **plus** the pending turn) for inspection, whereas `$get_turns()` / `$set_turns()` operate on history only (ellmer re-appends the pending turn) — including the warning not to feed `turns` back into `$set_turns()`. `$on_request_end()` now documents the cancelled/interrupted case (it still fires; `turn` may be an `AssistantPartialTurn` with `NA` `@tokens`/`@cost`).

On tests: one rewrites history from inside `on_request_start()` and asserts the *next* request reflects it (the compaction contract); another asserts both callbacks fire on the `$chat_async()` loop. NEWS and the PR description are updated to match.

Thanks for the thorough review!

gadenbuie (approved):
This looks good to me! I pushed a few small changes:

1. Updated the documentation to let users know that `on_request_end()` only fires when the turn ends partially or successfully. As in, it won't fire when an unhandled error occurs that brings down the chat lifecycle. I think this is okay; devs can use `tryCatch()` or `$catch()` as needed to explicitly handle errors and the machinery there is better suited -- we would need to otherwise modify the request end callback to receive the error, which is awkward.

2. I updated some of the live-recorded tests to use dedicated VCR cassettes.

R/chat.R
```diff
@@ -506,6 +508,64 @@ Chat <- R6::R6Class(
     #' @return A function that can be called to remove the callback.
     on_tool_result = function(callback) {
       private$callback_on_tool_result$add(callback)
+    },
+
+    #' @description Register a callback that fires at the **start of each model
```
hadley: @thisisnic can you please trim these docs a bit?

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* `Chat` gains `$on_request_start(turns)` and `$on_request_end(turn)`, paired callbacks that fire before and after each model request in the tool loop (once per request, whether or not it involves tools). `$on_request_start()` receives the **complete outgoing request** (history plus the pending turn) for inspection -- e.g. to measure its size and decide whether to compact; rewrite the conversation with `$set_turns()`, which operates on history (the canonical idiom being `chat$set_turns(compact(chat$get_turns()))`). `$on_request_end()` receives the assistant turn just returned, useful for per-request latency/cost tracking and for observing tool-call requests before they run (#1051).
```
hadley: @thisisnic and this could be shorter too.

hadley (approved):

PR #1053, opened 2026-07-06
Extract out Model from Provider

### LINKED ISSUE #499: Resolve the `Provider` design problem
Currently it includes both information about the provider (i.e. the base url and API key) and the model (e.g. model name and parameters).

thisisnic:
This PR creates a new Model class to properly separate the concepts of models/providers.

Closes #499 

Summary of changes:
- Provider class loses `model`, `params` and `extra_args` properties which now instead belong to Model
- lots of methods now take model object as a parameter
- `batch_chat()` calculates hash based on model as well as other pre-existing fields


R/chat.R
```diff
@@ -95,9 +102,14 @@ Chat <- R6::R6Class(
...
+    #' @description Retrieve the model name.
     get_model = function() {
-      private$provider@model
+      private$model@name
+    },
+
+    #' @description Retrieve the Model object. For expert use only.
+    get_model_object = function() {
+      private$model
     },
```
thisisnic, line 138: I don't love the name of this and it breaks symmetry with `get_provider()` which returns the provider object and not the name.

However, if we refactor `get_model()` to return the model object, it's a breaking change - although we don't export the `Chat` class (yet), we do allow access to it via `chat()`.

thisisnic:
I did a quick Claude analysis to see which of our revdeps will likely by affected (we'll need to run a proper revdepcheck to confirm though) and results are below.  FYI @gadenbuie, @simonpcouch, and @cpsievert (but I'm happy to make PRs to the relevant repos!)
______________________________________________________________________

:robot::

| Package | Hits | What breaks |
|---------|------|-------------|
| **btw** | 3 | `provider@model` read/write + `get_provider()@model` in test |
| **lang** | 2 | `provider@model` read + `x@model <-` assignment in test helper |
| **mcptools** | 1 | `ellmer::Provider("dummy", "dummy", "dummy")` - old 3-arg constructor |
| **mini007** | 1 | `meta_data@model` - needs investigation to confirm `meta_data` is an ellmer Provider |
| **querychat** | 2 | `ellmer::Provider("test", "test", "test")` in tests |
| **shidashi** | 2 | `provider@model` + `ellmer::Provider(name = "dummy", ...)` |
| **vitals** | 4 | `get_provider()@params$max_tokens` (4 places) |

7 of 43 revdeps may be affected. Searched for `@model`, `@params`, and `@extra_args` access on provider objects, `ellmer::Provider()` construction with old args, and direct `Chat$new()` calls (excluding subclasses like `QueryChat$new()`).

simonpcouch:
Thanks for the proactive look at revdeps + writeup on the actual resulting issues in those packages, this was helpful to see. 

I haven't looked carefully at the code and don't have strong reflexes for the implications for ellmer internals here, but this seems reasonable to me. If this makes yall's work easier, I'm happy to put the work in downstream (though given your question on the linked issue about whether the original serialization motivation still applies, noting that this does seem like it'd be a good bit of work).

I don't see any evidence that this is actually where yall are headed, but putting this out there for the GH archeologists: I had initially wondered whether the intent here was to eventually attach additional metadata to the Model class, like context length, pricing, capabilities, API flavor. (For open-weights models especially,) The provider decides context length, price, and the API shape, so if ellmer eventually gets in the business of pushing that metadata around, it'd need to continue being observed at the Model x Provider intersection (like `prices` is).

R/chat.R
gadenbuie: It's a bit of a departure from the current patterns, but I wonder if it's worth considering making `$model` an active binding? That would not only give direct access to the model object, but would also let users modify in the model settings in place, e.g.

```r
chat$model@temperature <- 0.7
```

I can justify the pattern change, to myself at least, by rationalizing provider as something you are not expected to change for a `Chat` instance, while the new `Model` object contains settings that you _would_ want to be able to change without having to create an entirely new chat.

gadenbuie:
Re revdeps, it looks to me like

* Most `@model` calls could be `$get_model()` on the chat instance
* Most `@model <-` assignments are now covered by `$set_model()`
* Maybe `Provider()` could be special cased to promote scalar `model` strings to `Model` instances, which would cover most of the `Provider("test", "test")` uses.
* Similarly, maybe we can have `@params` become a getter/setter that reaches into `@model` to paper over the difference (with a deprecation warning, and the setter would throw). I'm tempted to say that if only vitals uses this pattern, it'd be okay to directly break it, but I have a feeling this pattern might show up in user code. Maybe worth a GitHub search to estimate usage in the wild?

So overall, I think the breaking changes are worth it :)

thisisnic:
Thanks both for taking a look at this! :)

@simonpcouch - that's helpful to think of looking forward.  Totally agree re metadata, we should probably keep this `Model` class really lightweight and just reflect user config, and maybe we'd need some other concept to represent the model x provider intersection.  

@gadenbuie - I took a look and could only find 1 example that wasn't a fork of ellmer or vitals, so I'm thinking a clean break and submitting PRs to the affected repos might be the simplest way forwards.

R/chat.R
thisisnic: I can totally see how that could be a nice way for users to update things, though the one thing that's giving me pause is how it fits in with the existing ways of updating things. Will need to think more about this one!

hadley (approved):
LGTM!

PR #1059, opened 2026-07-14
Preserve OpenRouter error messages

xmarquez:
## Summary

- Preserve the provider-supplied message when OpenRouter returns an error object.
- Add a snapshot test for this response-error path.

## Root cause

`check_openrouter_error()` extracted `error$message`, but passed the literal string `"message"` to `abort()`. This masked the provider's actual error text.

## Validation

- Ran `air format` on the modified R files.
- Loaded the package and verified that the resulting condition preserves the provider message.

### DIFF
--- a/R/provider-openrouter.R
+++ b/R/provider-openrouter.R
@@ -141,7 +141,7 @@ check_openrouter_error <- function(error, call = caller_env()) {
   }
 
   abort(
-    c("message", i = if (!is.null(details)) details),
+    c(message, i = if (!is.null(details)) details),
     call = call
   )
 }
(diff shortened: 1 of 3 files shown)


xmarquez:
Hello @thisisnic - apologies if I needed to raise an issue first; this seemed like a very simple bug I ran into using openrouter for a project (provider error messages were not preserved). Let me know if I need to do anything differently.

hadley:
Doh.

thisisnic (approved):
Thanks very much for fixing this! No issue needed first :)

Issue #1063, opened 2026-07-22
Bedrock prices missing for all modern Claude models


hadley:
`data-raw/prices.R` maps only litellm's `bedrock` provider to `"AWS/Bedrock"`, but litellm splits Bedrock models across three providers. The `inner_join()` with `provider_lookup` silently drops the other two:

| litellm provider | models | mapped |
|---|---|---|
| `bedrock` | 267 | yes |
| `bedrock_converse` | 138 | **no** |
| `bedrock_mantle` | 13 | **no** |

Every modern Claude id lives under `bedrock_converse` (65 Claude ids there vs. 78 legacy ones under `bedrock`), so none of them reach the bundled table. The Claude rows we do ship are all legacy dated ids like `anthropic.claude-3-5-sonnet-20241022-v2:0`.

This includes ellmer's own default Bedrock model, so `token_usage()` and `$get_cost()` silently report `NA` out of the box:

```r
pkgload::load_all()
match_prices("AWS/Bedrock", c(
  "us.anthropic.claude-sonnet-4-6",           # chat_aws_bedrock() default
  "anthropic.claude-sonnet-5",
  "anthropic.claude-3-5-haiku-20241022-v1:0"  # legacy id, works
))
#> # A tibble: 3 × 3
#>   cached_input input output
#>          <dbl> <dbl>  <dbl>
#> 1        NA     NA       NA
#> 2        NA     NA       NA
#> 3         0.08   0.8      4
```

`anthropic.claude-opus-4-8` and anything matching `sonnet-4-6` are likewise absent.

## Fix

Add `bedrock_converse` and `bedrock_mantle` to `provider_lookup`, both mapping to `"AWS/Bedrock"`. Two things to watch:

- Model ids may collide across the three litellm providers, so the bind needs deduplicating.
- The `stopifnot()` row-count guards will need new thresholds.

Mapping `bedrock_mantle` also picks up pricing for the mantle-only models (GPT-5.4/5.5/5.6, Grok 4.3); all 13 have `input_cost_per_token` and `output_cost_per_token` upstream.

hadley:
@thisisnic can you take this please?

thisisnic:
Will do!

PR #1064, opened 2026-07-22
Add bedrock-mantle support to chat_aws_bedrock()

hadley:
AWS now serves Bedrock models from two endpoints. The original `bedrock-runtime` endpoint speaks the Converse API, which is what ellmer has always used. The newer `bedrock-mantle` endpoint speaks the Anthropic Messages and OpenAI Responses APIs instead, and hosts models that Converse can't reach at all — including the GPT-5 family, Grok 4.3, and Claude Mythos. Until now those models were simply unavailable through ellmer.

`chat_aws_bedrock()` gains an `api` argument to choose between them. You rarely need to set it: ellmer picks the API from the model name and falls back to Converse for anything it doesn't recognise, so existing code behaves exactly as before.

## Example

```r
# Frontier OpenAI models, which only exist on bedrock-mantle
chat <- chat_aws_bedrock(model = "openai.gpt-5.6-sol")
chat$chat("What is 1 + 1? Just the number.")
#> 2

# Claude through the Anthropic Messages API rather than Converse
chat <- chat_aws_bedrock(model = "anthropic.claude-haiku-4-5", api = "messages")
```

Passing a mantle-only model without an `api` used to fail with a bare "The provided model identifier is invalid". That error now suggests the fix.

## Design decisions

**The argument names the request format, not the endpoint.** `api` takes `"converse"`, `"messages"`, or `"responses"` rather than `"runtime"` and `"mantle"`, because one mantle endpoint serves three different request formats — naming the endpoint wouldn't say which one to send.

**Chat Completions is deliberately not supported.** Every Bedrock model that speaks it also speaks either Converse or Responses, so it would add no model coverage.

**The model lookup is generated, not hand-written.** `data-raw/prices.R` already fetches litellm's model metadata weekly, and that data records which models live only on mantle. It now also writes `R/provider-aws-api.R`, so the table refreshes on the existing schedule. It's a plain R file rather than binary data in `sysdata.rda` so the weekly diff is reviewable.

**The table only lists models Converse can't serve.** Everything else falls through to Converse. This keeps a data refresh from silently moving a working call onto a different endpoint with different quotas — it can only ever turn a failure into a success. It also means an unfamiliar model id (a custom ARN, a new inference profile) keeps working as it does today rather than being rejected.

## Not included

* Bedrock pricing data is missing for all modern Claude ids, so `token_usage()` reports `NA` for them, including on mantle. That's a pre-existing bug in the same extraction script, tracked in #1063.
* Claude Mythos is listed as mantle-only in AWS's docs but doesn't appear in our account's model listing, so those two entries are doc-derived and untested.

## Verification

Live-tested against a real AWS account: basic chat, streaming, tool calling, and structured data all work on `openai.gpt-5.6-sol` via `"responses"`, and on `anthropic.claude-haiku-4-5` via `"messages"`.

That testing caught a bug worth flagging for review: mantle has two OpenAI-compatible paths that are disjoint rather than aliases. Newer models are served from `/openai/v1` and older open-weight ones like gpt-oss from `/v1`, and each rejects the other's models. AWS's API compatibility tables don't record this distinction — only the individual model cards mention it — so the first implementation used `/v1` and failed on exactly the models this PR is meant to unlock.

Note that `test-provider-aws.R:229` currently fails. It's unrelated to this PR — it fails identically on `main`, and was previously invisible because it skips without AWS credentials.


hadley:
@gadenbuie mostly an FYI so you can see that I'm using the liteLLM data in another way 
@thisisnic probably not much you can do here without AWS access, but please let me know if you spot anything odd.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* `chat_aws_bedrock()` gains an `api` argument to select between the Converse API on the `bedrock-runtime` endpoint and the Anthropic Messages or OpenAI Responses APIs on the `bedrock-mantle` endpoint. This makes models that Converse can't serve, like Claude Mythos and the GPT-5 family, available on Bedrock. The API is picked from `model` by default, so you only need to set `api` for models ellmer doesn't recognise.
```
```suggestion
* `chat_aws_bedrock()` gains an `api` argument to select between the Converse API on the `bedrock-runtime` endpoint and the Anthropic Messages or OpenAI Responses APIs on the `bedrock-mantle` endpoint. This makes models that Converse can't serve, like Claude Mythos and the GPT-5 family, available on Bedrock. The API is picked from `model` by default, so you only need to set `api` for models ellmer doesn't recognise (#1064).
```

R/provider-claude.R
```diff
@@ -166,17 +166,25 @@ method(base_request, ProviderAnthropic) <- function(provider) {
...
   req
 }
 
+method(base_request_error, ProviderAnthropic) <- function(provider, req) {
+  req_error(req, body = anthropic_error_body)
+}
+
+# <https://docs.anthropic.com/en/api/errors>
+anthropic_error_body <- function(resp) {
+  if (resp_content_type(resp) == "application/json") {
```
thisisnic:
```suggestion
  if (identical(resp_content_type(resp), "application/json")) {
```

R/provider-openai-compatible.R
```diff
@@ -162,20 +162,22 @@ method(base_request_error, ProviderOpenAICompatible) <- function(
...
-      } else {
-        prettify(resp_body_string(resp))
-      }
-    } else if (resp_content_type(resp) == "text/plain") {
-      resp_body_string(resp)
+  req_error(req, body = openai_error_body)
+}
+
+openai_error_body <- function(resp) {
+  if (resp_content_type(resp) == "application/json") {
```
thisisnic:
```suggestion
  if (identical(resp_content_type(resp), "application/json")) {
```

R/provider-aws.R
```diff
@@ -14,6 +16,30 @@ NULL
...
+#'   Claude Mythos) that Converse does not serve at all.
+#' * `"responses"` uses the OpenAI Responses API on the `bedrock-mantle`
+#'   endpoint. This is the only way to reach the GPT-5 family and Grok on
+#'   Bedrock. Note that mantle serves newer models from `/openai/v1` and older
+#'   open-weight models like gpt-oss from `/v1`; ellmer uses the former, so
+#'   reaching the latter needs an explicit `base_url`. They're all available
+#'   through `"converse"` anyway.
+#'
+#' By default ellmer picks the API from `model`, falling back to `"converse"`
+#' for models it doesn't recognise. Set `api` explicitly to override this.
```
thisisnic:
```suggestion
#' for models it doesn't recognize. Set `api` explicitly to override this.
```

Reluctantly making this suggestion :joy:

R/provider-aws.R
```diff
@@ -14,6 +16,30 @@ NULL
 #' [Claude](https://aws.amazon.com/bedrock/claude/), using the Bedrock
 #' [Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html).
```
thisisnic: Should we mention the other APIs above?

thisisnic (changes requested):
I think we might also need to override `has_batch_support()` and `count_tokens()` for `ProviderAWSBedrockMessages` and `ProviderAWSBedrockResponses` as their parents use specific URLS which don't make sense here

thisisnic:
@hadley I'm gonna push some changes to this branch and merge it as it's blocking some other PRs I want to review and merge ahead of the v 0.5.0 release, so it might be worth you reviewing the changes I made to see if we want to revert anything in a follow-up PR.

The most obvious thing I wasn't sure about was the change I made in https://github.com/tidyverse/ellmer/pull/1064/changes/a01dc1d6017560a1473bb3acae323c67b89363d8 as now the error message triggers on a broader set of circumstances.

PR #1065, opened 2026-07-27
Include bedrock_converse and bedrock_mantle in pricing data

### LINKED ISSUE #1063: Bedrock prices missing for all modern Claude models

thisisnic:
All of the below summary was written by Claude, but I manually checked the changes to the file by running the script before/after, and also ran the example in the reprex from the issue to confirm that the correct prices were coming through.

## Summary

- Adds `bedrock_converse` and `bedrock_mantle` to the litellm provider lookup, both mapping to `"AWS/Bedrock"`. These were silently dropped by the `inner_join()`, so modern Claude models (including ellmer's default `us.anthropic.claude-sonnet-4-6`) had no pricing data and `token_usage()` / `$get_cost()` reported `NA`.
- Deduplicates after the join since a small number of models appear under multiple litellm providers with negligible rounding differences.
- Bumps the `stopifnot()` row threshold from 500 to 1000 to guard against accidentally losing these rows in future.

Closes #1063.

## Test plan

- [x] Run `Rscript data-raw/prices.R` and verify it completes without error
- [x] Compare before/after pricing: only AWS/Bedrock rows added, no unexpected removals
- [x] Verify `match_prices("AWS/Bedrock", "us.anthropic.claude-sonnet-4-6")` returns prices instead of `NA`


hadley:
Thanks!

PR #1073, opened 2026-08-02
Update error messages based on whether connecting to ollama via IP address or locally

### LINKED ISSUE #886: Clarify model checking in `chat_ollama()`
Maybe only check the model when working locally? Otherwise you get a confusing message as in #875.

thisisnic:
Fixes #886


hadley (approved):
Makes sense to me

PR #1075, opened 2026-08-04
Explore contents_trajectory(): serialize chats to trajectory-v1

gadenbuie:
## Summary

This is an exploratory draft. The implementation is in solid shape, but there's no pressure to merge this — it's here to spark discussion about whether this fits ellmer's scope. It's inspired by [letta-ai/trajectory](https://github.com/letta-ai/trajectory), a schema/toolkit for normalizing raw agent transcripts (Claude Code, Codex, etc.) into a common `trajectory-v1` JSON shape. ellmer already holds a fully structured `Chat`/`Round`/`Turn`/`Content` object graph in memory, so the parsing problem that project solves doesn't apply here — this PR asks a narrower question: does ellmer's internal model map cleanly onto that shape, and is a native R serializer worth having?

`contents_trajectory()` is a new generic, added alongside the existing `contents_text()` / `contents_html()` / `contents_markdown()` family (same S7 dispatch pattern), that converts a `Chat`, `Round`, or `Turn` into a plain R list matching `trajectory-v1`'s shape: an ordered stream of `user` / `assistant` / `reasoning` / `tool` records preceded by one `meta` record. Callers serialize to JSON themselves via `jsonlite::toJSON()`.

Key design decisions (see inline docs and commit messages for detail):
- No dependency on the TS/Node `@letta-ai/trajectory` package — this is a from-scratch reshaping of data ellmer already has, not a transcript parser.
- Timestamps are omitted when ellmer has no absolute time for a turn (only OpenAI's Responses API reports one), rather than synthesizing fake ones.
- Content types with no `trajectory-v1` role appearing directly in a turn (images, PDFs, uploaded files) degrade to a short placeholder string (e.g. `[inline image]`) rather than being silently dropped.
- A tool result whose value is itself `Content` (e.g. an image returned by a screenshot tool) is a different case — that's real tool output, not decoration — so it's inlined as a base64 data URI instead of a placeholder, and instead of crashing or requiring wire-protocol-style expansion. The same base64-inline approach could be extended to images/PDFs appearing directly in a turn too, if placeholder-only turns out to be too lossy in practice.
- Validated in tests against a locally-vendored, timestamp-relaxed copy of the upstream JSON Schema (test-only, no new runtime dependency).
- Upstream added an optional `ok: boolean` field to `tool` records after this PR was started (a source-native success/failure signal, e.g. Claude Code's `is_error`). ellmer has an exact match for that signal in `ContentToolResult@error`, so `tool` records now carry `ok` for regular tool calls; built-in web search/fetch results have no equivalent signal and omit it, matching upstream's "omit when unreliable" policy.

## Verification

```r
chat <- chat_openai()
chat$chat("What's 2 + 2?")

trajectory <- contents_trajectory(chat)
jsonlite::toJSON(trajectory, auto_unbox = TRUE, null = "null", pretty = TRUE)
```

See `tests/testthat/test-content-trajectory.R` for coverage of tool calls/results, reasoning content, mixed-content assistant turns, built-in web search/fetch tools, and `Round`-level conversion.


simonpcouch:
If you're game, let's pair on this review! Would be helpful to have some back and forth on the shape of this. :)

PR #1081, opened 2026-08-10
Update dates in tests to fix failures on main

thisisnic:
Some of the tests in #968 had dates for comparison which at the time were in the future but now are in the past, and so we are seeing failures on main, so I've updated them to be in the distant future and we'll worry about this again when ellmer is 100 years old :grinning:

### DIFF
--- a/tests/testthat/test-prices.R
+++ b/tests/testthat/test-prices.R
@@ -40,8 +40,6 @@ test_that("prices_get() uses a newer cached snapshot", {
 
   df <- prices[1, ]
   df$input <- 9999
-  # Far-future date: must stay newer than the bundled snapshot, which the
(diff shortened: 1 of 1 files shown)


tests/testthat/test-prices.R
```diff
@@ -40,10 +40,12 @@ test_that("prices_get() uses a newer cached snapshot", {
 
   df <- prices[1, ]
   df$input <- 9999
+  # Far-future date: must stay newer than the bundled snapshot, which the
+  # update-prices workflow refreshes regularly
```
thisisnic:
```suggestion
```

gadenbuie:
🤦 Thanks for fixing this!

PR #1082, opened 2026-08-11
Support output schemas in tool definitions

alesanGreat:
## Summary

Adds an optional `output_schema` argument to `tool()` and stores it on `ToolDef` as a `TypeObject`.

This enables downstream MCP servers such as `posit-dev/mcptools` to advertise the MCP 2025-06-18 `outputSchema` field without changing tool execution behavior. It also preserves positional `ToolDef()` compatibility by appending the new property after the existing properties.

Related downstream issue: https://github.com/posit-dev/mcptools/issues/103

## Validation

- `tests/testthat/test-tools-def.R` passes with the new storage/type-check tests.
- Verified existing positional `ToolDef(..., convert, annotations)` construction still maps arguments correctly and leaves `output_schema = NULL`.
- `git diff --check` passes.
- A broader local suite run reached the affected `tools-def` tests successfully; unrelated provider tests in the isolated container require optional packages and credentials that are not configured there.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@
 * `chat_openrouter()` now correctly preserves provider error messages (@xmarquez, #1059).
 * New `models_update_prices()` downloads the latest model pricing data from GitHub and saves it to a local cache. Subsequent calls to `token_usage()` and related functions will use the updated prices (#968).
 * New `Model` class separates model configuration (name, parameters, extra arguments) from the `Provider` class, which now only captures API endpoint details. `Chat` gains a new `$get_model_object()` method to retrieve the `Model` object. This is a breaking change for anyone who directly accesses `provider@model`, `provider@params`, or `provider@extra_args`; use the `Model` object instead (@thisisnic, #499).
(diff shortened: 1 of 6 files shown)


thisisnic:
Thanks for the PR @alesanGreat.  I took a look, and had a think and I don't think we're going to end up merging this as the output schema is not used within ellmer and might add a bit of confusion about the API surface as something only useful for mcptools, and so is a bit too niche to become an extra argument on the `tool()` function.

I also had a look at which providers support it, and it appears to only be supported by Google Gemini but none of the other main ones - we tend to add features only once they're widely adopted to ensure we're keep things straightforward and standardised.

alesanGreat:
Thanks for the thoughtful review. That API boundary makes sense for ellmer if output_schema is not broadly useful across its own providers yet. I’ll close this proposal rather than keep a direction open that the project does not want to carry, and I’ll explore whether the downstream mcptools use case can be solved without expanding ellmer’s public 	ool() surface. Thanks again for taking the time to evaluate it.

PR #1086, opened 2026-08-13
Always include `parameters` in Databricks tool definitions

### LINKED ISSUE #1084: Error Using Tool With Databricks As Model Provider
Hi Ellmer Team, 

I am running into an error that says Bad request: "parameters" is a required property for "function" object in parameter "tools" when I ran querychat along with ellmer using databricks model provider. The following is the code that I ran that can be used to reproduce the error. Is it happening due to ellmer::tool since querychat uses it under the hood (i.e. https://github.com/posit-dev/querychat/blob/bfd59f636af33778b291fcaa590ca385126188a5/pkg-r/R/querychat_tools.R#L29) ? 

I am using a free version of Databricks and have a personal access token generated as mentioned under the section Create personal access tokens for workspace users at this link (https://docs.databricks.com/aws/en/dev-tools/auth/pat). 

Please help! @thisisnic 

library(querychat)
library(palmerpenguins)

databricks_client <- ellmer::chat_databricks(
    workspace = "https://dbc-6daebf39-7148.cloud.databricks.com",
    model = "databricks-meta-llama-3-3-70b-instruct",
    token = "secret_key"
)

querychat_app(penguins, client = databricks_client)

Warning: Error in req_perform_connection: HTTP 400 Bad Request.
ℹ Bad request: "parameters" is a required property for "function" object in parameter "tools"

Thanks so much! 

Best, 
Kola

thisisnic:
Fixes #1084 

Registering a tool with no arguments caused `chat_databricks()` to error as we omitted the `parameters` field entirely from the request when tools have no arguments, as a precaution in #549, which doesn't appear to be needed any more.  

I compared against other models with tools registered with/without parameters and all ran find after this fix. It also brings us in line with chatlas.

### DIFF
--- a/R/provider-databricks.R
+++ b/R/provider-databricks.R
@@ -235,27 +235,6 @@ method(base_request_error, ProviderDatabricks) <- function(provider, req) {
   })
 }
 
-# See: https://docs.databricks.com/aws/en/machine-learning/foundation-model-apis/api-reference#functionobject
-method(as_json, list(ProviderDatabricks, ToolDef)) <- function(
-  provider,
-  x,
-  ...
-) {
-  # Note: It seems that Databricks doesn't support the "strict" field, despite
-  # what their documentation says. It *is* supported for structured outputs,
-  # though. I suspect a copy & paste error in their docs.
-  compact(list(
-    type = "function",
-    "function" = compact(list(
-      name = x@name,
-      description = x@description,
-      # Use the same parameter encoding as the OpenAI provider. Databricks
-      # requires `parameters` even when the tool has no arguments (#1084).
-      parameters = as_json(provider, x@arguments, ...)
-    ))
-  ))
-}
-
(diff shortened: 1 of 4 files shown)


R/provider-databricks.R
```diff
@@ -249,11 +249,9 @@ method(as_json, list(ProviderDatabricks, ToolDef)) <- function(
     "function" = compact(list(
       name = x@name,
       description = x@description,
-      # Use the same parameter encoding as the OpenAI provider, but only if
-      # there actually are parameters.
-      parameters = if (length(x@arguments@properties) != 0) {
-        as_json(provider, x@arguments, ...)
-      }
+      # Use the same parameter encoding as the OpenAI provider. Databricks
```
hadley: Can we now drop this entire method?

hadley (approved):

R/provider-databricks.R
thisisnic: Oh yeah, of course!!

PR #1089, opened 2026-08-13
Report the model's context window in `models_*()` and on `Chat`

### LINKED ISSUE #1083: Model metadata omits the context window
The `models_*()` functions don't report how much context a model accepts, and nothing on `Chat` exposes it either. A caller can see how many tokens a conversation has accumulated, through `get_tokens()`, but not how close that is to the limit — leaving a hardcoded per-model table that goes stale with every release, or per-provider HTTP calls, as the only ways to find out. That is the work ellmer otherwise saves callers from doing.

Most providers already return it, in a response ellmer parses and then discards. The table below is a brief overview gathered from their docs, so it may well have shifted by the time anyone acts on it.

| Provider | Endpoint | Field |
| --- | --- | --- |
| Anthropic | `/v1/models` | `max_input_tokens`, `max_tokens` |
| OpenRouter | `/api/v1/models` | `context_length` |
| Gemini | `/models` | `inputTokenLimit`, `outputTokenLimit` |
| Mistral | `/v1/models` | `max_context_length` |
| vLLM | `/v1/models` | `max_model_len` |
| LM Studio | `/v1/models` | `max_context_length` |
| Ollama | `api/show` | `<arch>.context_length`, not in the `api/tags` ellmer calls |
| OpenAI | `/v1/models` | Absent |
| Bedrock | `ListFoundationModels` | Absent |
| Portkey | `/models` | Absent; normalised to the OpenAI shape before it reaches ellmer |

Five of those are a parser widening — the Anthropic parser keeps `id`, `display_name` and `created_at`, and the vLLM and LM Studio parsers keep `id` alone. Ollama needs a second call. For the last three there is nothing to read, which is what `NA` is for. OpenAI's model object is only `id`, `object`, `created` and `owned_by`, so everyone keeps their own table — including OpenAI, whose Codex CLI [added a `max_context_window` field to its own model metadata](https://github.com/openai/codex/pull/18382) in April.

Suggestion: add it to the `models_*()` data frames as `max_input_tokens`, or whatever name generalises best, `NA` where the provider doesn't report it. Exposing it on `Chat` too would pair naturally with `get_tokens()`, and would give the compaction hook in #1051 the number it needs to decide.

Happy to send a PR if you'd like one.

nbenn:
Most providers publish a model's context window in a response `models_list()` already parses and then discards. This adds a `context_window` column to the `models_*()` data frames, and `Chat$get_context_window()` to pair with `$get_tokens()`.

| Provider | Field read |
| --- | --- |
| Anthropic | `max_input_tokens` |
| Gemini | `inputTokenLimit` |
| Groq | `context_window` |
| LM Studio | `max_context_length`, from the native `/api/v1/models` |
| Mistral | `max_context_length` |
| Posit | `max_context_length` |
| vLLM | `max_model_len` |
| Ollama | `<arch>.context_length`, from the `api/show` response already fetched for `capabilities` |

Everyone else gets `NA`, so the column is there whichever provider you ask. The OpenAI-compatible parser reads `context_window` generically, which is what covers Groq.

The `Chat$get_context_window()` method resolves the current model through `models_list()`, caching per provider, base URL and model, so it costs one request per session.

Verified live: Gemini 36/36 models, Mistral 55/55, Groq 15/15, Ollama Cloud 18/18, Posit 10/10, LM Studio against a local server. Anthropic comes from the existing `anthropic-list-models` cassette and vLLM from vLLM's source. Vertex publishes no token limit anywhere in `publisherModels`, in either projection, so it falls through to `NA`. For LM Studio, its OpenAI-compatible `/v1/models` carries only `id`, `object` and `owned_by`, so the window is joined from the native `/api/v1/models` by model key, and a server without that route degrades to `NA`. The `test_models()` helper now checks the column, and checks it is populated for the providers that report one.

Fixes #1083

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # ellmer (development version)
 
 * ellmer now preserves provider citations in chat history and streamed content, and displays cited sources in console output. Enable cited web results with `chat$register_tool()`, using tools such as `openai_tool_web_search()`, `claude_tool_web_search()`, or `claude_tool_web_fetch(citations = TRUE)` (#1068).
+* `Chat` gains a `$get_context_window()` method that reports how many tokens the current model's context window holds, so you can tell how much of it `$get_tokens()` has used (@nbenn, #1083).
 * `Chat` gains a `$token_count()` method that estimates the number of tokens in new input using the provider's token counting endpoint (@thisisnic, #814).
 * `chat_anthropic()`, `chat_aws_bedrock()`, and `chat_posit()` now default to `claude-sonnet-5`. `chat_openai()` and `chat_openrouter()` now default to `gpt-5.6-terra` (@thisisnic, #1066).
 * `chat_anthropic()` now correctly handles the `fallback` content block returned when a model's server-side refusal fallback (`server-side-fallback-2026-06-01`) is triggered (@simonpcouch, #1058).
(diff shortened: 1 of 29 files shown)


gadenbuie:
> This adds a `context_window` column to the `models_*()` data frames, and `Chat$get_context_window()` to pair with `$get_tokens()`.

Thanks for preparing the PR @nbenn! I have two high-level comments just from reading the PR description:

1. It seems like `max_context_length` is the most common name for the field, and I personally prefer it. It reads as the most obvious and clear name for the field. I think the method should be named similarly.
2. Also, after #1053, we now have a separate `Model` object, like `Provider`, that can be accessed through `Chat$get_model_object()` (IIRC). I think `$get_max_content_length()` would be better positioned as a method on that class (or maybe even a `max_content_length` active property, but we should follow precedent there). This way, the max content length is attached to the upstream model.

nbenn:
Thanks — happy to rename to `max_context_length`, and it is the most common spelling among the fields this reads: three providers use it verbatim (Mistral, Posit, LM Studio) and Ollama uses `context_length`, against a single `context_window` (Groq). For the record on why we landed where we did, `"context_window"` is already ellmer's own word for the concept — `provider-claude.R` and `provider-aws.R` both map `model_context_window_exceeded` onto a `finish_reason` of `"context_window"`, documented in `R/turns.R` — so the rename does mean naming one thing two ways. Fine by me either way.

Happy to move the accessor onto `Model` too. Mostly I want to check we're aligned on what that implies, since #1053 deliberately landed the other way: @simonpcouch [noted](https://github.com/tidyverse/ellmer/pull/1053#issuecomment-5132264146) that the provider decides context length, so it "would need to continue being observed at the Model x Provider intersection (like `prices` is)", and @thisisnic [agreed](https://github.com/tidyverse/ellmer/pull/1053#issuecomment-5159396284) the class should stay "really lightweight and just reflect user config".

That matters here because the value genuinely is provider-dependent. The same model reports a different window depending on who serves it:

| | Anthropic direct | Posit gateway |
| --- | --- | --- |
| `claude-sonnet-4-5` | 1000000 | 200000 |
| `claude-sonnet-4-6` | 1000000 | 1000000 |

So a `Model` can only carry the right number once a provider is in scope, which is the constraint any shape has to satisfy. How would you like it handled? I'm glad to hang it off `Model`, or to go further and add the intersection concept @thisisnic floated, whichever you'd prefer.

PR #1095, opened 2026-08-15
Implement provider files API

### LINKED ISSUE #1091: Implement provider files API
See https://github.com/posit-dev/chatlas/pull/334 for the equivalent in Chatlas

thisisnic:
Fixes #1091.

Upload a file to the provider once and reference it by id, instead of re-sending its bytes on every turn:

```r
chat <- chat_anthropic()

report <- chat$file_upload("quarterly-report.pdf")
chat$chat(report, "Summarize the key findings.")
chat$chat("Now list the risks.")   # no re-upload

chat$file_list()
chat$file_get(report)
chat$file_delete(report)
```

Five experimental `Chat` methods (`$file_upload()`, `$file_list()`, `$file_get()`, `$file_download()`, `$file_delete()`) wrapping internal S7 generics that dispatch on provider. Supported for `chat_openai()`, `chat_anthropic()`, and `chat_google_gemini()`; other providers error. `ContentUploaded` is exported and gains `provider` and `extra`.

Also fixes bugs in the existing code: Anthropic's `container_upload` block had the wrong shape, unrecognized MIME types produced an invalid block, `claude_file_list()` only returned the first page, and timestamps were parsed with bare `as.POSIXct()`.

## Things to review

* **`claude_file_*()` and `google_upload()` are superseded, not deprecated.** They're now thin wrappers over the new generics and keep working without warnings, since they're the only way to manage files without first building a chat (cf. `models_*()`). Happy to deprecate or remove instead.
* **Anthropic's Files API went GA on 2026-08-19**, after this branch started. This PR targets the GA API (no beta header, `next_page` pagination, `expires_at` populated).


R/chat.R
```diff
@@ -263,6 +263,78 @@ Chat <- R6::R6Class(
...
+    #'   `ContentUploaded(uri = "gs://bucket/object", mime_type = ...)`.
+    #' * An OpenAI upload can also be referenced from a
+    #'   [chat_openai_compatible()] chat pointed at OpenAI's Chat Completions
+    #'   API, except for images, which that API can't reference by id.
+    #' @param path Path to a file to upload.
+    #' @param mime_type MIME type of the file. If not supplied, it's guessed
+    #'   from the file extension.
+    #' @return A [ContentUploaded] that can be passed to `$chat()` and
+    #'   friends in place of the file itself.
+    file_upload = function(path, mime_type = NULL) {
```
hadley: Should we have a `temporary` argument where it gets delete when the Chat object is finalized?

R/content-pdf.R
hadley, line 4: Probably need docs update; will be better to recommend explicitly uploading the file.

R/files.R
```diff
@@ -0,0 +1,85 @@
...
+  "provider",
+  function(provider, id, ...) {
+    S7_dispatch()
+  }
+)
+method(file_delete, Provider) <- function(provider, id, ...) {
+  no_file_support(provider)
+}
+
+no_file_support <- function(provider, call = caller_env()) {
```
hadley, line 58: Nice!

R/provider-google-upload.R
```diff
@@ -22,9 +19,14 @@
...
 #' @examples
 #' \dontrun{
+#' # Old:
 #' file <- google_upload("path/to/file.pdf")
+#' chat <- chat_google_gemini()
+#' chat$chat(file, "Give me a three paragraph summary of this PDF")
 #'
+#' # New:
 #' chat <- chat_google_gemini()
+#' file <- chat$file_upload("path/to/file.pdf")
```
thisisnic: Note to self: no need to have both, just show the new one

hadley (approved):
Worth be worth checking out if this overlaps at all with this sort of prompt:

```R
chat <- chat_claude()
chat$chat("Write me a brief story")
chat$chat("Turn that into a markdown artifact")
```

Or is the artefact just a tool provided only in the web chat Ui?

If it does overlap, it would be nice to provide an example somewhere.

Also I think it's fine to deprecate the existing functions since they clutter the API.

thisisnic:
> Worth be worth checking out if this overlaps at all with this sort of prompt:
> 
> ```r
> chat <- chat_claude()
> chat$chat("Write me a brief story")
> chat$chat("Turn that into a markdown artifact")
> ```
> 
> Or is the artefact just a tool provided only in the web chat Ui?
> 
> If it does overlap, it would be nice to provide an example somewhere.

Had a look into it and looks like it's only in the web chat UI unfortunately.

R/content-pdf.R
thisisnic: Done

R/chat.R
thisisnic: Had a look at ways of implementing this but it felt super messy in terms of things that could go wrong with finalizers (e.g. a session crashes etc) and cloned `Chat` objects, so I figure instead we could just use the `expires_in`/`expires_at` with a default value of 48 hours - hopefully this cover what you're after here, but if not let me know and I'll update it.

Issue #376, opened 2025-03-19
Bad Request on next chat if Claude model on Bedrock returns an empty response


jvroberts:
After a few queries using Claude 3.7 Sonnet model on AWS Bedrock, it occasionally responds to something with an empty response (usually as part of a tool call, but could be otherwise).  Can repro with the following contrived example: 

```
# connect to Claude 3.7 model via Bedrock
chat <- chat_bedrock(
  model = "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
)

chat$chat("Respond with only three blank lines")
chat$chat("This will fail")
```
Error in `req_perform_connection()`:
! HTTP 400 Bad Request.
• The text field in the ContentBlock object at messages.1.content.0 is blank. Add text to the text field, and try again.

hadley:
Same problem with claude directly:

``` r
library(ellmer)

chat <- chat_claude()
#> Using model = "claude-3-7-sonnet-latest".
chat$chat("Respond with only three blank lines")
chat$chat("This will fail")
#> Error in `req_perform_connection()`:
#> ! HTTP 400 Bad Request.
#> • messages: text content blocks must contain non-whitespace text [invalid_request_error]
```

<sup>Created on 2025-03-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hadley:
I'm not 100% sure there's anything I can do here. I'm already respecting the documented API and it feels like a relatively rare special case.

hadley:
Actually @wch figured out a nice way to solve this; PR incoming shortly.

etbrand:
I'm seeing this issue with Claude on Databricks too:

``` r
chat <- ellmer::chat_databricks(model = "databricks-claude-3-7-sonnet")
chat$chat("Respond with only three blank lines")
chat$chat("This will fail")
Error in `req_perform_connection()`:
! HTTP 400 Bad Request.
ℹ {"message":"messages: text content blocks must contain non-whitespace text"}
Run `rlang::last_trace()` to see where the error occurred.
```

<sup>Created on 2025-11-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1099, opened 2026-08-19
Handle empty responses in AWS and Claude

### LINKED ISSUE #1070: chat_aws_bedrock() fails after empty response from claude-sonnet-5
With `claude-sonnet-5` on Bedrock, asking the model to respond with only blank lines can produce an assistant turn with an empty thinking block and no text content. The empty thinking block is dropped by `as_json(ProviderAWSBedrock, ContentThinking)` (returns `NULL` when thinking is `""`), and `compact()` strips it, leaving `content = list()`. On the next turn, Bedrock rejects the empty content array with HTTP 400.

The fix from #376 / PR #393 handles whitespace-only `ContentText`, but this is a different case: the content array is completely empty because the only content block (empty thinking) was dropped.

## Reprex

```r
library(ellmer)
chat <- chat_aws_bedrock(model = "us.anthropic.claude-sonnet-5")
chat$chat("Respond with only two blank lines")
chat$chat("What's 1+1? Just give me the number")
#> Error in `req_perform()`: HTTP 400 Bad Request.
#> The content field in the Message object at messages.1 is empty.
#> Add a ContentBlock object to the content field and try again.
```

The same issue likely applies to `chat_anthropic()` since its `as_json(ProviderClaude, ContentThinking)` has the same `return()` for empty thinking.

Found while working on #1066 (bumping default models to `claude-sonnet-5`).

thisisnic:
Fixes #1070

`claude-sonnet-5` sometimes returns a completely empty response (no content blocks). This broke things in two ways:

- `chat_anthropic()` errored when constructing the turn from the empty response.
- `chat_aws_bedrock()` sent the empty turn back as part of the conversation history on the next request, and Bedrock rejects empty messages with HTTP 400.

I checked whether the existing #376 workaround (replacing whitespace-only text with `[empty string]`) could be removed but it can't: both APIs still reject whitespace-only text and this is a different case, where there's no text at all.

The fix that :robot: came up with is that when an assistant turn has no content, both providers now send a single `[empty string]` text block in its place. I had :robot: try dropping the turn entirely (which is what `chat_anthropic()` did before, from #711), but that leaves two user messages in a row and the model tends to treat them as one, so the earlier instruction carries over into the later question. With the placeholder, the model answers the follow-up question correctly most of the time. I verified the #711 tool-calling scenario still works on both providers.

### DIFF
--- a/R/provider-claude.R
+++ b/R/provider-claude.R
@@ -777,7 +777,10 @@ method(as_json, list(ProviderAnthropic, Turn)) <- function(
   } else if (is_user_turn(x) || is_assistant_turn(x)) {
     x <- turn_contents_expand(x)
     content <- as_json(provider, x@contents, ...)
-    if (is_assistant_turn(x) && length(content) == 0) {
+    if (length(content) == 0) {
+      if (!is_assistant_turn(x)) {
+        return()
+      }
       # Dropping empty assistant turns confuses the model, so send a
       # placeholder instead (#711, #1070)
       content <- list(list(type = "text", text = "[empty string]"))
--- a/R/provider-claude.R
+++ b/R/provider-claude.R
@@ -530,6 +530,7 @@ method(value_turn_with_turns, ProviderAnthropic) <- function(
       )
     }
   }))
+  contents <- contents %||% list()
 
   tokens <- value_tokens(provider, result)
   cache_write <- result$usage$cache_creation_input_tokens %||% 0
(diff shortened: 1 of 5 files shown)


R/provider-claude.R
```diff
@@ -750,15 +751,15 @@ method(as_json, list(ProviderAnthropic, Turn)) <- function(
...
-    if (is_assistant_turn(x) && identical(x@contents, list())) {
-      # Drop empty assistant turns to avoid an API error
-      # (all messages must have non-empty content)
-      return(NULL)
-    }
     x <- turn_contents_expand(x)
     content <- as_json(provider, x@contents, ...)
     if (length(content) == 0) {
-      return(NULL)
+      if (!is_assistant_turn(x)) {
```
hadley, line 781: Can we use the same logic as bedrock? Or is this subtly different?

hadley (approved):

R/provider-claude.R
thisisnic: Nah, this was a mistake here, updated to match

PR #1102, opened 2026-08-20
feat: stream structured output

cpsievert:
## Why this matters
`Chat$stream()` and `Chat$stream_async()` can now stream provider-native structured output as it is generated, enabling callers to render JSON-backed artifacts and other structured responses incrementally. This follows the public API and orchestration precedent from [Chatlas PR #262](https://github.com/posit-dev/chatlas/pull/262).

## What changes
- Add `type` support to synchronous and asynchronous streaming.
- Preserve raw JSON fragments in text mode and `Content` chunks in content mode, while saving the completed response as `ContentJson`.
- Suppress registered tools for structured requests and reject provider/model/type combinations that require tool-based structured output.
- Cover OpenAI root-schema wrapping, Anthropic and Bedrock fallback rejection, tool behavior, and streaming modes.

## Verification
- `Rscript -e 'devtools::test(".", filter = "chat")'` (`614` passing)\n- `air format R/chat.R R/chat-structured.R tests/testthat/test-chat.R`\n- `devtools::document()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # ellmer (development version)
 
 * ellmer now preserves provider citations in chat history and streamed content, and displays cited sources in console output. Enable cited web results with `chat$register_tool()`, using tools such as `openai_tool_web_search()`, `claude_tool_web_search()`, or `claude_tool_web_fetch(citations = TRUE)` (#1068).
+* `Chat$stream()` and `Chat$stream_async()` now suppress registered tools when given a `type_()` specification and store the completed assistant turn as `ContentJson`; streaming structured output requires native provider support and does not use tool-based fallback.
 * `Chat` gains a `$token_count()` method that estimates the number of tokens in new input using the provider's token counting endpoint (@thisisnic, #814).
 * `chat_anthropic()`, `chat_aws_bedrock()`, and `chat_posit()` now default to `claude-sonnet-5`. `chat_openai()` and `chat_openrouter()` now default to `gpt-5.6-terra` (@thisisnic, #1066).
 * `chat_anthropic()` now correctly handles the `fallback` content block returned when a model's server-side refusal fallback (`server-side-fallback-2026-06-01`) is triggered (@simonpcouch, #1058).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # ellmer (development version)
 
 * ellmer now preserves provider citations in chat history and streamed content, and displays cited sources in console output. Enable cited web results with `chat$register_tool()`, using tools such as `openai_tool_web_search()`, `claude_tool_web_search()`, or `claude_tool_web_fetch(citations = TRUE)` (#1068).
-* `Chat$stream()` and `Chat$stream_async()` now suppress registered tools when given a `type_()` specification and store the completed assistant turn as `ContentJson`; streaming structured output requires native provider support and does not use tool-based fallback.
+* `Chat$stream()` and `Chat$stream_async()` gain a `type` argument for streaming structured output from providers with native support (@cpsievert, #1102)
 * `Chat` gains a `$token_count()` method that estimates the number of tokens in new input using the provider's token counting endpoint (@thisisnic, #814).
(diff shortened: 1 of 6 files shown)


cpsievert:
@thisisnic I'm wanting this for an upcoming [querychat feature](https://github.com/posit-dev/querychat/pull/245). If you're planning on a release soonish, would be great if we could fit this in.

thisisnic:
> @thisisnic I'm wanting this for an upcoming [querychat feature](https://github.com/posit-dev/querychat/pull/245). If you're planning on a release soonish, would be great if we could fit this in.

Sounds good, I'll add it to the 0.5.0 milestone

R/chat.R
```diff
@@ -650,10 +693,20 @@ Chat <- R6::R6Class(
...
       yield_as_content = FALSE,
       controller = NULL
     ) {
+      if (
+        !is.null(type) &&
+          uses_tool_structured_output(private$provider, private$model, type)
+      ) {
+        cli::cli_abort(
+          "Streaming structured output requires native provider support for the supplied model."
+        )
```
thisisnic:
```suggestion
cli::cli_abort(c(
    "Can't stream structured output with {private$provider@name} model {.val {private$model@name}}.",
    i = "Streaming requires native structured output, but this provider and model fall back to tool calling.",
    i = "Use `$chat_structured()` instead."
  ))
```

How about something like this to give the user a way forward?

NEWS.md
```diff
@@ -1,6 +1,7 @@
 # ellmer (development version)
 
 * ellmer now preserves provider citations in chat history and streamed content, and displays cited sources in console output. Enable cited web results with `chat$register_tool()`, using tools such as `openai_tool_web_search()`, `claude_tool_web_search()`, or `claude_tool_web_fetch(citations = TRUE)` (#1068).
+* `Chat$stream()` and `Chat$stream_async()` now suppress registered tools when given a `type_()` specification and store the completed assistant turn as `ContentJson`; streaming structured output requires native provider support and does not use tool-based fallback.
```
```suggestion
* `Chat$stream()` and `Chat$stream_async()` gain a `type` argument for streaming structured output from providers with native support (@cpsievert, #1102)
```

tests/testthat/test-chat.R
```diff
@@ -335,6 +335,529 @@ test_that("can perform a simple async batch chat", {
...
+          input_tokens = 1,
+          output_tokens = 2,
+          input_tokens_details = list(cached_tokens = 0)
+        )
+      )
+    )
+  )
+}
+
+test_that("stream() supports native structured output", {
```
thisisnic, line 419: This looks good but the testing is a bit more expansive than for other features, and perhaps we could trim it a little?

:robot:'s suggestion:

> Could we collapse to the `run_case(async)` pattern used elsewhere in this file, with one factory that takes a vector of deltas? I think five tests would cover it: native happy path, `stream = "content"`, scalar wrapping, tools omitted, and all the rejection cases in a single `expect_snapshot(error = TRUE)`.

thisisnic (commented):
Just a few changes to suggest from me, otherwise looks good!

hadley (approved):

R/chat.R
cpsievert: Done — adopted your suggested message (with provider/model name and the `$chat_structured()` pointer) at both the sync and async abort sites: https://github.com/tidyverse/ellmer/commit/36af76a2fccfb0a929a4c1635bbdbaf2a202d77b

NEWS.md
cpsievert: Done — replaced with your suggested bullet: https://github.com/tidyverse/ellmer/commit/36af76a2fccfb0a929a4c1635bbdbaf2a202d77b

tests/testthat/test-chat.R
cpsievert: Done — consolidated to a `run_case(async)` pattern with a single delta-based response factory. The rejection cases now live in one test using `expect_snapshot(error = TRUE)` (the mock still errors if a request starts, preserving the before-request check). One deviation: the Bedrock scenario is guarded with `tryCatch(skip = ...)` since `chat_aws_bedrock_test()` skips without AWS credentials, which would otherwise abort the whole test: https://github.com/tidyverse/ellmer/commit/36af76a2fccfb0a929a4c1635bbdbaf2a202d77b

PR #1104, opened 2026-08-21
Default base_url to `ANTHROPIC_BASE_URL` and `AWS_ENDPOINT_URL_BEDROCK_RUNTIME`

### LINKED ISSUE #1103: Respect `ANTHROPIC_BASE_URL` and `AWS_ENDPOINT_URL_BEDROCK_RUNTIME`
The official Anthropic and AWS SDKs recognize environment variables `ANTHROPIC_BASE_URL` and `AWS_ENDPOINT_URL_BEDROCK_RUNTIME`. In ellmer these base URLs are fixed unless you pass `base_url` yourself.

It would be nice if `base_url` could default to these variables when they're set, matching the SDKs, and fall back to the public endpoints otherwise. That way someone pointing anthropic or bedrock at a proxy or gateway sets the variable once and ellmer picks it up like, instead of passing `base_url` by hand to every `chat_*()` and `models_*()` call.

karawoo:
Fixes #1103.

`chat_anthropic()` and `chat_aws_bedrock()` now default `base_url` to the `ANTHROPIC_BASE_URL` and `AWS_ENDPOINT_URL_BEDROCK_RUNTIME` environment variables when they're set, matching the official SDKs. Setting the variable is enough to route anthropic or bedrock through a proxy or gateway, so you don't have to pass `base_url` on every call.


hadley (approved):

thisisnic:
OK, I've merged in main and made a few additional changes to incorporate #1064.  The only significant behavioural change I made here was having `models_aws_bedrock()` work with the `AWS_ENDPOINT_URL_BEDROCK` rather than the runtime var.

PR #1105, opened 2026-08-21
forward viewer session token to Connect's LLM gateway

karawoo:
When chat traffic on Posit Connect goes through Connect's LLM gateway, this forwards the viewer's session token as a `Posit-Connect-User-Session-Token` header so Connect can attribute the call to the viewer. The header is added in `chat_perform()` and `parallel_turns()`. It's only added when running on Connect, and only sent to Connect's own gateway URL.

For now this is specific to shiny sessions running on Connect as it retrieves the user session token from the shiny session via `getDefaultReactiveDomain()`. I didn't attach the header in `batch_chat()` as my understanding is batch jobs, being async and resumed across sessions, wouldn't make sense to run in an individual user's shiny session. Let me know if that doesn't sound right, though.


hadley (approved):
Approving in principle; @thisisnic can you please take a look at the code too?

R/httr2.R
```diff
@@ -129,6 +130,47 @@ ellmer_req_robustify <- function(req, is_transient = NULL, after = NULL) {
...
+}
+
+connect_viewer_token <- function() {
+  # Only read the session token when actually running on Connect
+  if (!running_on_connect() || !is_installed("shiny")) {
+    return(NULL)
+  }
+  session <- shiny::getDefaultReactiveDomain()
+  session$request$HTTP_POSIT_CONNECT_USER_SESSION_TOKEN
+}
```
thisisnic: We've got the same function name here as `connectcreds::connect_viewer_token()` which we use in other places in ellmer.  Perhaps we want to rename this to something like `connect_session_token()`?

R/httr2.R
```diff
@@ -129,6 +130,47 @@ ellmer_req_robustify <- function(req, is_transient = NULL, after = NULL) {
...
+  server <- Sys.getenv("CONNECT_SERVER")
+  if (identical(server, "")) {
+    return(FALSE)
+  }
+  server <- url_parse(server)
+  url <- url_parse(url)
+  identical(tolower(url$scheme), tolower(server$scheme)) &&
+    identical(tolower(url$hostname), tolower(server$hostname)) &&
+    identical(url$port, server$port) &&
+    startsWith(url$path %||% "", "/__gateway__/")
```
thisisnic: :robot: had this feedback about subpath URLs but I wasn't sure if it's right or not, as I don't have enough context - perhaps this will make more sense to you though @karawoo?

> `CONNECT_SERVER` for running content is derived from that server address (it's the same variable connectcreds uses to reach Connect's API), so on such a deployment it would be https://proxy.company.com/positconnect/. If the gateway lives on the Connect server, its URLs would then be `https://proxy.company.com/positconnect/__gateway__/...` — whose parsed path is `/positconnect/__gateway__/...`, which fails `startsWith(url$path, "/__gateway__/")`. Result: the token is never attached and viewer attribution silently doesn't work for every sub-path deployment. 
> The fix is small if it's needed: anchor the check to the server's own path instead of the root, something like:
> `prefix <- sub("/+$", "", server$path %||% "")`
> `startsWith(url$path %||% "", paste0(prefix, "/__gateway__/"))`
> which reduces to the current behaviour when `CONNECT_SERVER` has no path.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # ellmer (development version)
 
+* When running on Posit Connect, ellmer now forwards the viewer's session token to Connect's LLM gateway so gateway usage can be attributed to the viewer. This happens automatically for Shiny content and only affects requests to the gateway.
```
```suggestion
* When running on Posit Connect, ellmer now forwards the viewer's session token to Connect's LLM gateway so gateway usage can be attributed to the viewer. This happens automatically for Shiny content and only affects requests to the gateway (@karawoo, #1105).
```

thisisnic (approved):
A few minor points from me but otherwise looks good to go!  @karawoo - I made some changes to the accompanying PR for the env vars before I merged that one, so perhaps it's worth rebasing this one and checking you're happy everything works as you expect before I merge this PR?

R/httr2.R
karawoo: yeah that makes sense, I've updated it

tests/testthat/test-provider-aws.R
```diff
@@ -237,7 +237,7 @@ test_that("continues to work after whitespace only outputs (#376)", {
 test_that("aws_bedrock_api() guesses the api from the model", {
   expect_equal(aws_bedrock_api("us.anthropic.claude-sonnet-4-6"), "converse")
   expect_equal(aws_bedrock_api("anthropic.claude-mythos-5"), "messages")
-  expect_equal(aws_bedrock_api("openai.gpt-5.6-sol"), "responses")
+  expect_equal(aws_bedrock_api("openai.gpt-5.4"), "responses")
```
karawoo, line 240: this is unrelated to my change but was needed to get tests passing after https://github.com/tidyverse/ellmer/commit/a64f94e644718c0598b01b0cd50a3c21c2646435

thisisnic: Thanks for letting us know, have opened https://github.com/tidyverse/ellmer/issues/1110

PR #1106, opened 2026-08-25
Add `conversation_id` active binding to `Chat`

cpsievert:
Part of posit-dev/shinychat#307. Python equivalent: posit-dev/chatlas#398. Downstream consumer: posit-dev/shinychat#343.

## Summary

Adds a `conversation_id` active binding to `Chat`. When set, the identifier is recorded as the `gen_ai.conversation.id` attribute on the `invoke_agent` and `chat` OpenTelemetry spans emitted for subsequent model calls, so backends can group spans belonging to the same conversation (per the [OTel GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)).

This is developer-facing: intended for frameworks that manage conversation history rather than end users. Because ellmer owns the spans, downstream frameworks don't need to define any OpenTelemetry spans of their own — they just hand over the conversation ID. In shinychat's case (posit-dev/shinychat#343), this drops its entire OTel span apparatus (tracer resolution, promise-domain span wrapping, otel/otelsdk dependencies, span tests) in favor of a single assignment before the model call.

Following the conventions, ellmer never generates an identifier on its own — when unset (`NULL`, the default), the attribute is omitted entirely rather than falling back to an invented value. Setting the binding only affects spans created for subsequent model calls; in-flight responses keep the ID captured at their submission.

## Verification

```r
chat <- chat_openai(model = "gpt-4o-mini")
chat$conversation_id <- "conv_123"  # recorded as gen_ai.conversation.id on chat/invoke_agent spans
chat$conversation_id <- NULL        # clear (e.g., starting a new conversation)
```

Tests in `tests/testthat/test-otel.R` cover: attribute present on both span types when set, absent by default, and binding round-trip/validation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,5 @@
 # ellmer (development version)
 
-* `Chat` gains a settable `$conversation_id` active binding. When set, the identifier is recorded as the `gen_ai.conversation.id` attribute on the OpenTelemetry `chat` and `invoke_agent` spans, allowing backends to group spans belonging to the same conversation (per the OpenTelemetry GenAI semantic conventions). Developer-facing: intended for frameworks that manage conversation history; ellmer never generates an identifier on its own.
-
 * ellmer now preserves provider citations in chat history and streamed content, and displays cited sources in console output. Enable cited web results with `chat$register_tool()`, using tools such as `openai_tool_web_search()`, `claude_tool_web_search()`, or `claude_tool_web_fetch(citations = TRUE)` (#1068).
 * `Chat` gains a `$token_count()` method that estimates the number of tokens in new input using the provider's token counting endpoint (@thisisnic, #814).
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # ellmer (development version)
 
+* `Chat` gains a settable `$conversation_id` active binding. When set, the identifier is recorded as the `gen_ai.conversation.id` attribute on the OpenTelemetry `chat` and `invoke_agent` spans, allowing backends to group spans belonging to the same conversation (per the OpenTelemetry GenAI semantic conventions). Developer-facing: intended for frameworks that manage conversation history; ellmer never generates an identifier on its own.
```
thisisnic: I'm wondering if we want to include this in `NEWS.md` or not, as it's a very developer-focused addition; what do you reckon?

cpsievert: Yep, good call, thanks. Done in f9adb71

cpsievert:
Ok if I merge @thisisnic?

PR #1107, opened 2026-08-26
Deprecate the model property instead of removing it

### LINKED ISSUE #1098: Properly deprecate `Provider@model`
`Provider@model` was removed rather than deprecated. We should properly deprecate it instead so existing code gets a lifecycle warning rather than an error.

S7 now supports deprecating properties, see https://github.com/RConsortium/S7/pull/731/changes#diff-52f22287badc71ac8c60e1b80950ef68c26dc58fce21847a85d40d640cbacfe9R310

thisisnic:
Fixes #1098 

We removed `Provider@model` in #1053 but we need to use a proper deprecation cycle for it so this PR introduces a warning instead of completely removing it immediately.

This feels super ugly, but I'm not sure there's currently a better way to do this?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,7 +13,7 @@
 * `chat_google_gemini()` no longer errors when mixing regular tools and built-in tools like `google_tool_web_search()` (@thisisnic, #1054).
 * `chat_openrouter()` now correctly preserves provider error messages (@xmarquez, #1059).
 * New `models_update_prices()` downloads the latest model pricing data from GitHub and saves it to a local cache. Subsequent calls to `token_usage()` and related functions will use the updated prices (#968).
-* New `Model` class separates model configuration (name, parameters, extra arguments) from the `Provider` class, which now only captures API endpoint details. `Chat` gains a new `$get_model_object()` method to retrieve the `Model` object. `provider@model`, `provider@params`, and `provider@extra_args` are deprecated and will be removed in a future release; use the `Model` object instead (@thisisnic, #499, #1098).
+* New `Model` class separates model configuration (name, parameters, extra arguments) from the `Provider` class, which now only captures API endpoint details. `Chat` gains a new `$get_model_object()` method to retrieve the `Model` object. `provider@model`, `provider@params`, and `provider@extra_args` are deprecated and will be removed in a future release; use the `Model` object instead (@thisisnic, #1098).
 * `tool()` functions that return complex objects like data frames or lists now produce a deprecation warning. Tool functions should return a character vector, an atomic vector, a JSON string (from `jsonlite::toJSON()`), or a `Content` object. Use `jsonlite::toJSON()` to convert complex objects before returning (@thisisnic, #858).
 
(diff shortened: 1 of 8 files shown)


R/provider.R
```diff
@@ -31,10 +33,30 @@ Provider <- new_class(
...
-    credentials = class_function | NULL
+    credentials = class_function | NULL,
+    # Deprecated: model details now live on Model. Remove after 0.5.0 (#1098).
+    model = prop_deprecated("model", "name"),
+    params = prop_deprecated("params", "params"),
+    extra_args = prop_deprecated("extra_args", "extra_args")
   )
 )
 
+# Default S7 print calls every getter, which would trigger the deprecation
```
hadley, line 44: Would you mind filing an issue upstream about this too?

thisisnic: Sure, done! 

https://github.com/RConsortium/S7/issues/754

tests/testthat/test-provider.R
```diff
@@ -22,3 +22,23 @@ test_that("models_list() dispatches through Chat to provider", {
...
+      base_url = "https://example.com",
+      model = "x"
+    )
+    provider@model
+  })
+})
+
+test_that("Provider print omits deprecated properties", {
+  expect_snapshot(print(test_provider()))
+})
```
thisisnic, line 44: Unsure whether to keep this or ditch it?

NEWS.md
```diff
@@ -13,7 +13,7 @@
...
 * `chat_openrouter()` now correctly preserves provider error messages (@xmarquez, #1059).
 * New `models_update_prices()` downloads the latest model pricing data from GitHub and saves it to a local cache. Subsequent calls to `token_usage()` and related functions will use the updated prices (#968).
-* New `Model` class separates model configuration (name, parameters, extra arguments) from the `Provider` class, which now only captures API endpoint details. `Chat` gains a new `$get_model_object()` method to retrieve the `Model` object. This is a breaking change for anyone who directly accesses `provider@model`, `provider@params`, or `provider@extra_args`; use the `Model` object instead (@thisisnic, #499).
+* New `Model` class separates model configuration (name, parameters, extra arguments) from the `Provider` class, which now only captures API endpoint details. `Chat` gains a new `$get_model_object()` method to retrieve the `Model` object. `provider@model`, `provider@params`, and `provider@extra_args` are deprecated and will be removed in a future release; use the `Model` object instead (@thisisnic, #499, #1098).
```
```suggestion
* New `Model` class separates model configuration (name, parameters, extra arguments) from the `Provider` class, which now only captures API endpoint details. `Chat` gains a new `$get_model_object()` method to retrieve the `Model` object. `provider@model`, `provider@params`, and `provider@extra_args` are deprecated and will be removed in a future release; use the `Model` object instead (@thisisnic, #1098).
```

tests/testthat/test-provider.R
hadley: Keep it, I think?

hadley (approved):

PR #1108, opened 2026-08-26
Don't send the unsupported `strict` field in Databricks tool definitions

SokolovAnatoliy:
## The problem

`chat_databricks()` sends `strict = TRUE` in tool definitions again, which Databricks does not support. Once enough tools are registered, **every** request fails with:

```
HTTP 400 Bad Request
Compiled grammar size (486.7MB) exceeds maximum allowed size (300MB)
```

This is a regression of #548. #1086 removed the `as_json(list(ProviderDatabricks, ToolDef))` method so that `parameters` is always sent (fixing #1084), but removing it made `ProviderDatabricks` inherit `ProviderOpenAICompatible`'s encoding, which hardcodes `strict = TRUE`. The removed method existed only to omit that field, and said so in a comment that went with it:

> It seems that Databricks doesn't support the "strict" field, despite what their documentation says. It *is* supported for structured outputs, though. I suspect a copy & paste error in their docs.

## Why this wasn't caught

**The failure changed shape since #548 was filed.** Back then Databricks rejected the field outright (`tools.0.custom.strict: Extra inputs are not permitted`). It now *accepts* it and compiles the tool schema into a sampling grammar instead.

So `strict = TRUE` looks harmless on a small tool set — #1086's updated snapshot expecting `strict = TRUE` passes, and so does everything in the test suite. The cost is per model-choosable property against a 300 MB grammar budget, so only a large enough tool set crosses the threshold. And because grammar compilation happens *before* generation, once you cross it every request fails, not just the ones using the offending tool. An agent with 20 tools / 30 properties went from working to answering nothing, with no failing test anywhere.

Measured against a real 20-tool, 30-property set on `databricks-claude-sonnet-4-6`:

| Tool encoding | Result |
|---|---|
| as shipped (`strict = TRUE`) | HTTP 400, 486.7 MB grammar |
| `strict = FALSE` | answers |
| `strict` omitted | answers |

For what it's worth, tool count is not the axis: 30 trivial synthetic tools compile fine, while 20 real ones do not. `ProviderVllm` and `ProviderGroq` already omit `strict`, so Databricks joining them is consistent with existing practice.

## The fix

Restores the method, **keeping #1086's behaviour**: `parameters` is always included, including for tools with no arguments (#1084). The only difference from the inherited encoding is the absent `strict`.

Also adds a regression test asserting `strict` is absent and `parameters` present, for both a zero-argument and an argument-taking tool — so the behaviour is pinned by a test rather than resting on a comment that can be removed along with the method. The comment now explains the newer failure mode too.

`testthat::test_file("tests/testthat/test-provider-databricks.R")`: 29 passed, 0 failed, 4 skipped.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,5 @@
 # ellmer (development version)
 
-* `chat_databricks()` no longer sends the unsupported `strict` field in tool definitions. This fixes a regression introduced in #1086, where inheriting the OpenAI-compatible encoding made Databricks compile every tool schema into a sampling grammar; once enough tools were registered, all requests failed with `HTTP 400 Compiled grammar size (...MB) exceeds maximum allowed size (300MB)` (#548).
 * `chat_aws_bedrock()` gains an `api` argument to select between the Converse API on the `bedrock-runtime` endpoint and the Anthropic Messages or OpenAI Responses APIs on the `bedrock-mantle` endpoint. This makes models that Converse can't serve, like Claude Mythos and the GPT-5 family, available on Bedrock. The API is picked from `model` by default, so you only need to set `api` for models ellmer doesn't recognize (#1064).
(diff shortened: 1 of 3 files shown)


thisisnic:
Oops, good catch, thanks for reporting this and making the PR!

thisisnic (approved):

PR #1135, opened 2026-09-06
fix: keep optional tool args out of `required` for Posit models

gadenbuie:
## Summary

`chat_posit()` routes non-Claude models through `ProviderPositOpenAI`, which inherited OpenAI's strict-mode tool schema: every property listed in `required`, with optional ones made nullable. Gateway-hosted models that aren't OpenAI (Gemma, zai-org, etc.) don't follow that convention, so optional tool arguments were effectively mandatory.

`ProviderPositOpenAI` now has a `strict` property, set in `chat_posit()` from an `openai/` prefix on the model id. When `FALSE`, the provider uses the standard `TypeObject` serialization (optional args stay out of `required`), omits `strict` from tool definitions, and strips `strict` from structured-output `response_format` bodies. When `TRUE`, behavior is unchanged from the inherited OpenAI-compatible path.

The prefix check is a best guess about how the gateway might hypothetically name OpenAI-routed models; it's marked as such in a comment.

## Verification

```r
tool_def <- tool(
  function(x, y = 1) x + y,
  "Add numbers",
  arguments = list(
    x = type_number("First"),
    y = type_number("Second", required = FALSE)
  )
)
provider <- chat_posit(model = "google/gemma-4-26B-A4B-it")$get_provider()
as_json(provider, tool_def)$`function`$parameters$required
#> [[1]]
#> [1] "x"
```

`devtools::test(filter = "provider-posit")` passes (19 tests), including new coverage for both the strict and non-strict branches.


R/provider-posit.R
```diff
@@ -75,7 +75,10 @@ chat_posit <- function(
       name = "Posit",
       base_url = paste0(base_url, "/openai/v1"),
       extra_headers = api_headers,
-      credentials = credentials
+      credentials = credentials,
+      # Best guess: the gateway prefixes OpenAI-routed model ids with
+      # "openai/", but the catalog doesn't guarantee this yet.
+      strict = startsWith(model, "openai/")
```
gadenbuie, line 81: This prefix check is a best guess. The gateway catalog doesn't yet guarantee that OpenAI-routed models carry an `openai/` prefix, so if that naming never materializes, every Posit model takes the non-strict path. The goal is to avoid a need for future changes to this code path, but there's still a reasonable chance we'd need to revisit this if the catalog expands.

tests/testthat/test-provider-posit.R
```diff
@@ -46,6 +46,30 @@ test_that("gateway-specific errors get useful messages", {
...
+  tool_def <- tool(
+    function(x, y = 1) x + y,
+    "Add numbers",
+    arguments = list(
+      x = type_number("First"),
+      y = type_number("Second", required = FALSE)
+    )
+  )
+
+  provider <- chat_posit(model = "google/gemma-4-26B-A4B-it")$get_provider()
```
simonpcouch:
```suggestion
  provider <- chat_posit(model = "moonshotai/Kimi-K3")$get_provider()
```

Gemma 4 will probably be deprecated in the next week

R/provider-posit.R
simonpcouch: I can (try to) make this the reality if/when these get implemented :)

simonpcouch (approved):
Ah, right. This seems reasonable for now!

In the case that we do implement OpenAI routed models in the service, I think we would be using the responses API rather than the Completions API. If we're hoping that this "just works" if/when those models do make it into the service, we _could_ opt to send requests through a subclass of ProviderOpenAI rather than ProviderOpenAICompatible? My mental model isn't very strong on how ellmer handles that difference, though, and we can just cross that bridge when we come to it if we want, so reviewing as Approve.

tests/testthat/test-provider-posit.R
gadenbuie: This test doesn't actually call the API, and AFAICT we don't pre-validate the model for `chat_posit()`, so I don't think we even need to set the model.

```suggestion
  provider <- chat_posit()$get_provider()
```

gadenbuie:
I'm realizing this is probably a broader problem than just for `chat_posit()`. I think any provider that builds on the OpenAI-compatible API is likely to encounter the same problem. For example, I noticed today that a model called through `chat_lmstudio()` always wanted to include an optional argument, passing in `""` or another value that ends up serializing incorrectly in ellmer. It turned out that the parameter was included in `"required"`, which was throwing off the model.

tests/testthat/test-provider-posit.R
gadenbuie: oops, turns out this was here to make sure we use the OpenAI-compatible provider. Rather than test from `chat_posit()`, I just updated the test to use `PositOpenAIProvider` directly.

gadenbuie:
Following up on [my earlier comment](#issuecomment-5572812381): I investigated every provider that builds on `ProviderOpenAICompatible`, checking the ellmer source against each provider's own docs. Short version: the problem affects roughly eight more providers, and the fix belongs in the base class.

## What leaks

`ProviderOpenAICompatible` sends OpenAI's strict-mode convention in three independent places:

- `as_json(ToolDef)` adds `strict: true` to every tool definition
- `as_json(TypeObject)` marks **every** property `required` and makes optional ones nullable (`["number", "null"]`)
- `chat_body` adds `strict: true` to structured-output `response_format`

The all-required-plus-nullable convention is an OpenAI implementation detail of strict mode / constrained decoding ([OpenAI docs](https://platform.openai.com/docs/guides/structured-outputs)), not part of JSON Schema. It's only safe when the backend actually enforces it. Providers that do prompt-based tool calling put the schema in the prompt, and the model reads `required` the normal JSON-Schema way ("must fill these in"). A non-OpenAI model then treats optional args as mandatory, which is the Gemma bug here and the `chat_lmstudio()` bug I hit.

Because there are three separate pieces, providers can fix one and miss the others. Databricks removed the `strict` flag in #548 but never overrode `TypeObject`, so its schemas still declare every argument required. vLLM removed the tool-def flag but still sends `strict: true` in `response_format`. Cloudflare fixed tool serialization (Gemini-style) but not structured output.

## Per-provider findings

**Broken the same way as `chat_posit()` pre-fix** (inherit all three behaviors):

- **DeepSeek** — worst off. Strict tool calling is [beta-only](https://api-docs.deepseek.com/guides/tool_calls/) (requires `/beta` base URL), and the standard endpoint [only documents](https://api-docs.deepseek.com/api/create-chat-completion/) `text`/`json_object` response formats, so the `json_schema` + `strict` we send is unsupported. Stray `strict` fields can trigger schema-validation errors.
- **Mistral** — structured output [genuinely supports strict](https://platform-docs-public.pages.dev/capabilities/structured-output/custom_structured_output/), but its [function-calling docs](https://docs.mistral.ai/studio/conversations/function-calling) describe standard JSON Schema with no `strict` field on tool definitions. Optional tool args still effectively become required.
- **Perplexity** — docs describe the **opposite** convention: ["every declared property omitted from `required` remains optional"](https://docs.perplexity.ai/guides/structured-outputs). Our all-required schema directly conflicts with their documented semantics.
- **HuggingFace** — docs say [strict mode "is not supported by all providers"](https://huggingface.co/docs/inference-providers/guides/function-calling) and expose per-provider `supports_tools`/`supports_structured_output` metadata.
- **OpenRouter** — supports `strict` but warns that [enforcement varies by endpoint](https://openrouter.ai/docs/guides/features/structured-outputs); the default test model is OpenAI so we wouldn't notice.
- **Portkey** — pure passthrough to 3000+ models; [strict works for OpenAI upstreams](https://portkey.ai/docs/integrations/llms/structured-outputs), undefined otherwise.
- **LM Studio** — follows the OpenAI [tool format](https://lmstudio.ai/docs/developer/openai-compat/tools), but tool calling is prompt-based (llama.cpp grammar / Outlines is only for [structured output](https://lmstudio.ai/docs/developer/openai-compat/structured-output)), so nothing enforces the all-required convention. This is exactly the Gemma scenario.
- **Databricks** — as above: `strict` flag removed in #548, but schemas still all-required. Databricks' [function-calling docs](https://docs.databricks.com/aws/en/machine-learning/model-serving/function-calling) only support a JSON-Schema subset and don't list nullable type arrays; Databricks' own OpenAI client strips `strict` for non-GPT models.

**Half-fixed:**

- **vLLM** — tool-def `strict` removed, but `response_format` still carries `strict: true`, which [has caused validation errors](https://discuss.vllm.ai/t/the-openai-endpoint-doesnt-support-function-strict-setting/959) in some versions. (vLLM [recommends](https://docs.vllm.ai/en/stable/features/tool_calling/) the all-required-plus-nullable schema style for tool args, so the `TypeObject` serialization is actually fine there.)
- **Cloudflare** — tools use Gemini-style serialization (correct `required`), but structured output still sends `strict: true`; Cloudflare's [JSON mode](https://developers.cloudflare.com/workers-ai/features/json-mode/) "cannot guarantee" schema adherence.
- **Groq** — tools already patched. Groq's [structured outputs](https://console.groq.com/docs/structured-outputs) legitimately require all properties `required` with no optional fields, so strict is correct there.

**Fine as-is:**

- **Bedrock** — native Converse API, standard JSON Schema.
- **Snowflake Cortex** — all three behaviors overridden ([docs](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-rest-api)).
- **Ollama** — correct `required` convention via its own `TypeObject` override; a vestigial `strict: true` on tool defs is silently ignored upstream.
- **Azure / OpenAI** — real strict mode, [documented](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/structured-outputs), correct as-is.

## Suggested fix

Rather than patching eight providers individually, lift the `strict` property pattern from this PR into `ProviderOpenAICompatible` itself: default `FALSE`, gate the tool-def flag, the `TypeObject` serialization, and the `response_format` `strict` on it, and have Azure opt in (the only `ProviderOpenAICompatible` subclass serving genuine OpenAI models on the chat-completions API; `chat_openai()` uses the responses API and would not be affected). That fixes LM Studio, HuggingFace, OpenRouter, Portkey, DeepSeek, Mistral, Perplexity, Databricks, and vLLM at once. Groq needs care: its structured-output all-required requirement is real, so it would opt in for structured output while keeping the non-strict tool path.

PR #1139, opened 2026-09-07
`chat_posit()$set_model()` now switches model families

### LINKED ISSUE #1138: `chat_posit()` and `$set_model()` can't switch between Anthropic and OpenAI model families
### The problem

`chat_posit()` picks the provider class based on the model name. Claude models get a `ProviderPositAnthropic` pointing at `{base_url}/anthropic/v1`; everything else gets a `ProviderPositOpenAI` pointing at `{base_url}/openai/v1`.

https://github.com/tidyverse/ellmer/blob/d92a1b3e52a83048260b38d2c9350f1c48e1e2fb/R/provider-posit.R#L65-L80

This works fine at construction time. But `Chat$set_model()` only updates the model name. It never re-evaluates the provider choice. So when a Posit user switches to a model from the other family, they get an OpenAI-shaped request going to an Anthropic endpoint, or vice versa, which fails.

https://github.com/tidyverse/ellmer/blob/d92a1b3e52a83048260b38d2c9350f1c48e1e2fb/R/chat.R#L156-L161

For example:

```r
chat <- chat_posit(model = "claude-sonnet-5")
chat$chat("Hi")  # works
chat$set_model("zai-org/GLM-5.3-Flash")
chat$chat("Hi")  # fails: Anthropic-shaped request hits /openai/v1
```

This is unique to `chat_posit()`. Every other `chat_*()` function has a one-to-one mapping between its provider class and the set of models it can reach, so this bug can't happen anywhere else.

### Proposed fix: `provider_set_model()` hook

Add an S7 generic, `provider_set_model(provider, name)`, consulted by `Chat$set_model()` to optionally return an updated provider. The default method returns the provider unchanged, so no other provider is affected. `Chat$set_model()` gains one dispatch call:

```r
set_model = function(model) {
  check_string(model)
  private$model@name <- model
  private$provider <- provider_set_model(private$provider, model)
  provider_model(private$provider) <- private$model
  invisible(self)
}
```

`ProviderPositAnthropic` and `ProviderPositOpenAI` override it to swap in their sibling class, carrying over `credentials`, `extra_headers`, and the base URL (recoverable via `posit_gateway_url()`):

```r
method(provider_set_model, ProviderPositAnthropic) <- function(provider, name) {
  if (is_claude_model(name)) {
    provider
  } else {
    ProviderPositOpenAI(...)
  }
}
```

One design note: the `params()` argument of `chat_posit()` is already provider-agnostic. It uses standard names (`temperature`, `max_tokens`, ...) that each provider's `chat_params()` method translates to API-specific names, with a warning for unsupported parameters rather than an error. So the `Model` object carries over untouched when the provider swaps.

### Alternatives considered

The other idea was a unified `ProviderPosit` class holding both base URLs and picking the wire format per-request based on the current model name. S7 dispatches on an object's class, and a class can't change at runtime, so this would mean either:

1.  A `ProviderPosit` that internally delegates to an "active" Anthropic or OpenAI view, re-implementing every one of those family's ~30 methods as a thin wrapper. That includes all the `as_json(provider, Content*)` multi-dispatches, since every call site passes the provider object directly. Roughly 200 lines of boilerplate that can drift as methods are added upstream.
2.  Restructuring the provider architecture around a separate "wire format" concept that both families share. A much larger refactor that touches every provider.

The `provider_set_model()` hook gives users the same result with roughly 20 lines and no new abstraction. The only behavior note is that `cache` (Claude-only, lives on `ProviderPositAnthropic`) is silently lost when switching away from Claude models; the override should warn when that happens.

gadenbuie:
Fixes #1138

## Summary

`chat_posit()` picks its provider class from the model name: Claude models get `ProviderPositAnthropic` (pointing at `{base_url}/anthropic/v1`), everything else gets `ProviderPositOpenAI` (pointing at `{base_url}/openai/v1`). But `Chat$set_model()` only updated the model name, so switching to a model from the other family sent a request with the wrong wire format to the wrong endpoint.

This adds a `provider_set_model(provider, name)` S7 hook that `Chat$set_model()` consults. The default method returns the provider unchanged, so no other provider is affected. `ProviderPositAnthropic` and `ProviderPositOpenAI` override it to swap in their sibling class when the new model name belongs to the other family, carrying over credentials, `extra_headers`, and the gateway base URL (recovered via `posit_gateway_url()`).

Design notes:

- The `params()` argument of `chat_posit()` is provider-agnostic (each provider's `chat_params()` translates standard names to API-specific ones), so the `Model` object carries over untouched.
- The Claude-only `cache` setting survives family switches via an inert `cache` property on `ProviderPositOpenAI`, so `cache = "1h"` still applies after a round trip through OpenAI models. A chat that starts on an OpenAI model stores its `cache` argument too, so switching to Claude honors it. The property is hidden from `print()` for `ProviderPositOpenAI` since it's unused there.
- A unified `ProviderPosit` class that picks the wire format per request was considered, but S7 dispatches on class and it would need ~30 thin wrapper methods that can drift from upstream; the hook gives the same result in ~20 lines.

## Verification

```r
chat <- chat_posit(model = "claude-sonnet-5")
chat$chat("Hi")  # works

chat$set_model("zai-org/GLM-5.3-Flash")
chat$chat("Hi")  # now works: request goes to /openai/v1
```

New tests in `test-provider-posit.R` cover family swaps in both directions, no-op within a family, credentials/header carryover, cache preservation, and the hidden print field.

### DIFF
--- a/R/chat.R
+++ b/R/chat.R
@@ -156,6 +156,7 @@ Chat <- R6::R6Class(
     set_model = function(model) {
       check_string(model)
       private$model@name <- model
+      private$provider <- provider_set_model(private$provider, model)
       provider_model(private$provider) <- private$model
       invisible(self)
     },
--- a/R/provider-aws.R
+++ b/R/provider-aws.R
@@ -871,6 +871,12 @@ method(as_json, list(ProviderAWSBedrock, ContentThinking)) <- function(
     return()
   }
 
+  if (is.null(x@extra$signature) || !nzchar(x@extra$signature)) {
+    # Unsigned thinking blocks are rejected by Bedrock, so replay thinking
+    # from other providers as plain text.
+    return(as_json(provider, ContentText(format(x)), ...))
+  }
+
   list(
     reasoningContent = list(
       reasoningText = list(
--- a/R/provider-claude.R
+++ b/R/provider-claude.R
@@ -967,6 +967,12 @@ method(as_json, list(ProviderAnthropic, ContentThinking)) <- function(
     return()
   }
 
+  if (is.null(x@extra$signature) || !nzchar(x@extra$signature)) {
+    # Unsigned thinking blocks are rejected by the Anthropic API, so replay
+    # thinking from other providers as plain text.
+    return(as_json(provider, ContentText(format(x)), ...))
+  }
+
   list(
     type = "thinking",
     thinking = x@thinking,
--- a/R/provider-posit.R
+++ b/R/provider-posit.R
@@ -1,3 +1,4 @@
+#' @include provider.R
 #' @include provider-claude.R
 #' @include provider-openai-compatible.R
 #' @include files.R
@@ -75,7 +76,8 @@ chat_posit <- function(
       name = "Posit",
       base_url = paste0(base_url, "/openai/v1"),
       extra_headers = api_headers,
-      credentials = credentials
+      credentials = credentials,
+      cache = cache
     )
   }
   model <- Model(name = model, params = params, extra_args = api_args)
@@ -123,7 +125,12 @@ ProviderPositAnthropic <- new_class(
 
 ProviderPositOpenAI <- new_class(
   "ProviderPositOpenAI",
-  parent = ProviderOpenAICompatible
+  parent = ProviderOpenAICompatible,
+  properties = list(
+    # Inert on OpenAI models; carried over from/to ProviderPositAnthropic so
+    # it survives switching model families (#1138).
+    cache = prop_string(default = "5m")
+  )
 )
 
 method(base_request, ProviderPositAnthropic) <- function(provider) {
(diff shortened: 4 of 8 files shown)


simonpcouch (approved):
Nice catch!

cpsievert:
One follow-up issue this PR surfaces (found while mirroring it in chatlas, posit-dev/chatlas#430): after a family switch, the conversation history can contain `ContentThinking` blocks that were emitted by a non-Claude model via the OpenAI flavor (`provider-openai.R` wraps reasoning as `ContentThinking(extra = output)`, which has no Anthropic `signature`). When the chat then switches to a Claude model, `as_json(ProviderAnthropic, ContentThinking)` passes `x@extra$signature` straight through as `NULL`, and the Anthropic API rejects the request with `Invalid signature in thinking block`.

The fix we used on the Python side: when serializing turns for the Anthropic provider, drop `ContentThinking` blocks that lack a signature (and drop the turn entirely if nothing remains). Signed thinking from Claude itself is preserved. See `_is_replayable_thinking()` in posit-dev/chatlas#430 for the approach. Might be worth handling here or in a follow-up so the two implementations stay aligned.

gadenbuie:
> The fix we used on the Python side: when serializing turns for the Anthropic provider, drop `ContentThinking` blocks that lack a signature (and drop the turn entirely if nothing remains). Signed thinking from Claude itself is preserved. See `_is_replayable_thinking()` in [posit-dev/chatlas#430](https://github.com/posit-dev/chatlas/pull/430) for the approach. Might be worth handling here or in a follow-up so the two implementations stay aligned.

Great catch @cpsievert! I've encountered this before and ended up [solving it in btw](https://github.com/posit-dev/btw/blob/17c24685b614ba0336cc6ced0f43c6846332b63a/R/utils-ellmer.R#L145-L156) by replacing `ContentThinking` with `ContentText(format(content))` -- basically turning thinking blocks into text `<thinking>{content}</thinking>` blocks. That way they're carried into the new conversation, are generally visible to the new model, and show up in shinychat. It's not perfect but might be a slight improvement over dropping the content entirely.

gadenbuie:
Since Anthropic (or Anthropic models on Bedrock) is the only provider that does this, I also added the unsigned `ContentThinking` → `ContentText("<thinking>{text}</thinking>")` conversion in the Anthropic (and Bedrock) provider methods, too. That way anyone who takes turns from a non-Anthropic model for use as the turns on an Anthropic model will get the fix too.

PR #1142, opened 2026-09-10
Record time to first token on the OpenTelemetry chat span

schloerke:
Streamed responses now record the time to first token (seconds) as the `gen_ai.response.time_to_first_chunk` attribute on the `chat <model>` span.

- `R/otel.R`: `record_chat_otel_span_ttft_attr()` sets the attribute from a start timestamp captured just before the stream begins.
- `R/chat.R`: both sync and async streaming loops call it once, on the first non-empty text/thinking chunk.
- Tests: offline mocked test plus assertions in the live stream tests.

Follow-up in #1143 (stacked on this) broadens the trigger to tool calls and adds GenAI metrics.

### DIFF
--- a/R/chat.R
+++ b/R/chat.R
@@ -1005,7 +1005,7 @@ Chat <- R6::R6Class(
         acc$begin_turn(user_turn)
         on.exit(acc$finalize_turn(), add = TRUE)
 
-        stream_span <- NULL
+        stream_start <- Sys.time()
         result <- NULL
         for (chunk in response) {
           result <- stream_merge_chunks(private$provider, result, chunk)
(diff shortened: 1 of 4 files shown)


R/chat.R
```diff
@@ -1005,6 +1005,7 @@ Chat <- R6::R6Class(
         acc$begin_turn(user_turn)
         on.exit(acc$finalize_turn(), add = TRUE)
 
+        stream_start <- Sys.time()
```
shikokuchuo, line 1008:
```suggestion
        stream_start <- if (!is.null(chat_span)) Sys.time()
```

shikokuchuo: Also: the semconv measures this from request issuance, which happens inside `chat_perform()` (it blocks until headers arrive). Move this line above the `chat_perform()` call to match?

shikokuchuo (approved):

